PR #224, opened 2014-03-10
add opacity aesthetics

alyst:
I have not found alpha/opacity aesthetics in Gadfly. This patch should provide it (general definitions and Geom.Point implementation). In order for the patch to work, Compose.jl should be updated as well with another pull request (dcjones/Compose.jl/pull/45).


jiahao:
Superseded by db76d5c?

PR #516, opened 2014-12-19
Fix 509 : Guide.manual_color_key

aviks:


marcjwilliams1:
Hi, I'm still getting this error. I'm using julia v0.3.1 and the latest gadfly on ijulia. Do I need a newer version of julia for this fix to work? Thanks Marc

Issue #539, opened 2015-02-04
ERROR: `*` has no method matching *(::Date, ::Measure{Float64,MeasureNil})


IainNZ:
```
using Gadfly
using Dates

jl_date_vers = [Date(2014,08,20)  "v0.3.0";
                Date(2014,09,21)  "v0.3.1";
                Date(2014,10,21)  "v0.3.2";
                Date(2014,11,23)  "v0.3.3";
                Date(2014,12,26)  "v0.3.4";
                Date(2015,01,08)  "v0.3.5"]

p = plot(x=jl_date_vers[:,1], y=[10], label=jl_date_vers[:,2], Geom.label)
draw(SVG(8inch, 3inch), p)
```

Julia v0.3.4, Gadfly master

```
ERROR: `*` has no method matching *(::Date, ::Measure{Float64,MeasureNil})
 in deferred_label_context at /Users/idunning/.julia/v0.3/Gadfly/src/geom/label.jl:48
 in anonymous at /Users/idunning/.julia/v0.3/Gadfly/src/geom/label.jl:263
 in realize at /Users/idunning/.julia/v0.3/Compose/src/container.jl:319
 in drawpart at /Users/idunning/.julia/v0.3/Compose/src/container.jl:443
 in draw at /Users/idunning/.julia/v0.3/Compose/src/container.jl:400
 in draw at /Users/idunning/.julia/v0.3/Gadfly/src/Gadfly.jl:852
 in include at /usr/local/Cellar/julia/0.3.3/lib/julia/sys.dylib
 in include_from_node1 at loading.jl:128
 in process_options at /usr/local/Cellar/julia/0.3.3/lib/julia/sys.dylib
 in _start at /usr/local/Cellar/julia/0.3.3/lib/julia/sys.dylib (repeats 2 times)
while loading /Users/idunning/Dropbox/Websites/packages.julialang.org/script/breakgadfly.jl, in expression starting on line 12
```

IainNZ:
I think I have a "patch": in `label.jl`, change from
```julia
        x = Compose.absolute_x_position(x*cx, parent_transform, units,
                                        parent_box)
        y = Compose.absolute_y_position(y*cy, parent_transform, units,
                                        parent_box)
```

to
```julia
        x = Compose.absolute_x_position(Compose.x_measure(x), parent_transform, units,
                                        parent_box)
        y = Compose.absolute_y_position(Compose.y_measure(y), parent_transform, units,
                                        parent_box)
```

[...]

PR #540, opened 2015-02-04
Possible fix for #539

IainNZ:


dcjones:
Thanks! I think the issue was that I had to narrow that `*` function no avoid ambiguity warnings, and it ended up excluding dates. This does the trick though.

PR #554, opened 2015-02-27
WIP: Add tolerance for equality when optimizing ticks (Fix #550)

### LINKED ISSUE #550: Plot appears to hang when plotting certain vectors
The following code reliably hangs Gadfly:

```jl
data = repmat([sin(t/5) for t=1:100]', 10)'
S=svdfact(data) 
v=S[:V][:,1] #data matrix is of rank 1; this is the only nonsingular component
w=fill(S[:V][1,1], 10) #In exact arithmetic, equal to v
plot(x=1:10, y=w) #OK
plot(x=1:10, y=v) #Hangs

^CError showing value of type Plot:
ERROR: interrupt
 in optimize_ticks_typed at /Users/jiahao/.julia/v0.3/Gadfly/src/ticks.jl:115
 in optimize_ticks at /Users/jiahao/.julia/v0.3/Gadfly/src/ticks.jl:62
 in apply_statistic at /Users/jiahao/.julia/v0.3/Gadfly/src/statistics.jl:630
 in apply_statistics at /Users/jiahao/.julia/v0.3/Gadfly/src/statistics.jl:37
 in render at /Users/jiahao/.julia/v0.3/Gadfly/src/Gadfly.jl:717
 in display at /Users/jiahao/.julia/v0.3/Gadfly/src/Gadfly.jl:1003
 in display at /Users/jiahao/.julia/v0.3/Gadfly/src/Gadfly.jl:149
 in print_response at REPL.jl:139
 in print_response at REPL.jl:124
 in anonymous at REPL.jl:586
 in anonymous at REPL.jl:764
 in anonymous at LineEdit.jl:1199
 in prompt! at ./LineEdit.jl:1399
 in run_interface at ./LineEdit.jl:1374
 in run_frontend at ./REPL.jl:818
 in run_repl at ./REPL.jl:169
 in _start at ./client.jl:400
```

Note that `w` is almost equal to `v`, but with differences smaller than machine epsilon:

```jl
julia> w-v
10-element Array{Float64,1}:
 0.0        
 1.11022e-16
 0.0        
 0.0        
 0.0        
 0.0        
 0.0        
 0.0        
 0.0        
 5.55112e-17
```

darwindarak:
The main problem is in the [computation](https://github.com/dcjones/Gadfly.jl/blob/404f0aa433fa5431ba377234bd6d3427e9c174ab/src/ticks.jl#L93)

```julia
r = ceil((x_max - span) / (q*10.0^z * one_t))
```

where `z` is basically `log10(x_max - x_min)`.  If `x_max - x_min` is very small, `r` becomes too large.  Using `eps()` as the tolerance just happens to fix the failing example listed in #550, it is still too low for other examples.  For example:

```julia
y = ones(5)
y[1] = 1 + 2eps()

plot(x = 1:5, y = y, Geom.point)
```

will result in an infinite [loop](https://github.com/dcjones/Gadfly.jl/blob/404f0aa433fa5431ba377234bd6d3427e9c174ab/src/ticks.jl#L94)

@dcjones, any thoughts on how we can pick a reasonable tolerance?

### DIFF
--- a/src/ticks.jl
+++ b/src/ticks.jl
@@ -65,7 +65,7 @@ end
 function optimize_ticks_typed{T}(x_min::T, x_max::T, extend_ticks, Q::Vector{(Float64,Float64)}, k_min,
                            k_max, k_ideal, strict_span)
     one_t = one(T)
-    if x_min == x_max
+    if x_max - x_min < eps()*one_t
         R = typeof(1.0 * one_t)
         return R[x_min], x_min - one_t, x_min + one_t
     end


dcjones:
Thanks for looking into this. I wasn't able to figure out one epsilon that works, but I added another condition to preempt an infinite loop.

PR #581, opened 2015-04-11
hexbin color key title from "Color" to "Count"

visr:
This is copied from the `histogram2d` code. I believe it makes sense for `hexbin` as well.
http://gadflyjl.org/geom_histogram2d.html
http://gadflyjl.org/geom_hexbin.html


dcjones:
Yeah, that makes sense, thanks!

PR #582, opened 2015-04-12
Maximum number columns in key (legend) as a Theme parameter

wildart:
Sometimes, I need more columns in legend then 4, especially when legend is positioned horizontally.

Here is an example of the legend with a default setting of 4 columns:
key4

And the same plot with a maximum column number set to 12:
key12


dcjones:
Looks good, thanks!

PR #587, opened 2015-04-22
v0.4 updates (tuples, fieldnames)

### LINKED ISSUE #586: v0.4 Tuple changes cause failure to load
First error is in `Compose`, but likely are in other places

    ERROR: LoadError: LoadError: LoadError: invalid union type
    while loading .julia/v0.4/Compose/src/property.jl, in expression starting on line 453
    while loading .julia/v0.4/Compose/src/Compose.jl, in expression starting on line 61
    while loading .julia/v0.4/Gadfly/src/Gadfly.jl, in expression starting on line 6

Which are areas where Type unions are declared such as `Union(Nothing, (String,String))` instead of the new syntax `Union(Nothing, Tuple{String,String})`.

sebastiang:
Allows Gadfly to load, but tests don't pass. Attempts to fix #586

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -46,11 +46,6 @@ abstract GeometryElement    <: Element
 abstract GuideElement       <: Element
 abstract StatisticElement   <: Element
 
-# The names stashed on the Aesthetics object were retrieved with .names
-# in v0.3. That moves to fieldnames in v0.4. When v0.3 support is ended
-# you can delete this line
-if VERSION < v"0.4-dev" const fieldnames = names end
-
 
 include("misc.jl")
 include("ticks.jl")
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -46,6 +46,11 @@ abstract GeometryElement    <: Element
 abstract GuideElement       <: Element
 abstract StatisticElement   <: Element
 
+# The names stashed on the Aesthetics object were retrieved with .names
+# in v0.3. That moves to fieldnames in v0.4. When v0.3 support is ended
+# you can delete this line
+if VERSION < v"0.4-dev" const fieldnames = names end
+
 
 include("misc.jl")
 include("ticks.jl")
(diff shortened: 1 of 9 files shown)


sebastiang:
IT PLOTS! This revised PR allows a trivial plot to render in a browser. Note that this PR will only work for you if you're tracking master on a number of other packages; see related PRs. Fixing the tests is beyond my ken.

IainNZ:
Good stuff @sebastiang !

sebastiang:
This probably doesn't quite work unless you have the change in Color.jl. And now one starts to really wish for precompiled images :smiley:

sebastiang:
Found a few more tuples hiding in there and now more things plot. There are problems with legends, but at least most plots are running again. Fixes from here are probably going to have to come from @dcjones who actually knows how this stuff works!

I have to say the last set of errors were incredibly confusing. A call to `render` in `Gadfly.jl:818` was reporting `ERROR: TypeError: apply_type: in Array, expected Type{T}, got Tuple{DataType,DataType}`. But the problem was that `render` was actually calling another one, eventually down in `geom/line.jl:69` where there was some old-school `assq` style lists being created with `Array((Any,Any),1)`. Once fixed to the new syntax, things started working. But what a doozy of a message, so far removed from the source. I wonder if there is anything we can fix about the warning or the type inference to help out?

dcjones:
Thanks @sebastiang! I happened to have also just spent way too long trying to figure out the origin of that error. I tried with `--inline=no` also to no avail. I don't know why it would report such an incorrect call trace.

I cringed a bit when I saw I was making (potentially) big arrays of `(Any, Any)`, so I decided to rewrite to be a little more efficient. I'll have that pushed soon along with your fixes.

src/Gadfly.jl
```diff
@@ -46,6 +46,11 @@ abstract GeometryElement    <: Element
 abstract GuideElement       <: Element
 abstract StatisticElement   <: Element
 
+# The names stashed on the Aesthetics object were retrieved with .names
+# in v0.3. That moves to fieldnames in v0.4. When v0.3 support is ended
+# you can delete this line
+if VERSION < v"0.4-dev" const fieldnames = names end
```
tkelman: I think Compat should just handle this for you without needing this line

https://github.com/JuliaLang/Compat.jl/blob/2c91b7a12d1f206f06c902a2b687cc2587f7f614/src/Compat.jl#L195-L205

sebastiang: Thanks and good catch.

PR #591, opened 2015-05-05
Explicitly qualify `predict`

Keno:
`predict` is exported from both Loess and DataArrays which causes an error
on julia 0.4.


dcjones:
Thanks!

PR #609, opened 2015-06-20
Add documentation  Guide attributes in multi-layer plot

AlainLich:
Hi @dcjones 

This PR adds documentation for using Guide attributes (e.g. title) in a multi-layer plot. 

*It seems necessary to update the documentation in view of misleading information on various websites* like: 
http://stackoverflow.com/questions/21721344/how-to-add-legend-in-a-graph-when-using-package-gadfly-jl-in-julia


dcjones:
Thanks!

PR #631, opened 2015-07-08
Create guide_manual_color_key.md

amellnik:
Add a guide for Guide.manual_color_key since there wasn't already one.  I also snuck two handy things into the example which are not easily found in the guide:  modifying the color of an individual layer and what the name of the default "Gadfly blue" color is.  I'm not sure about the metadata at the top, the order might need to be changed.


dcjones:
Thanks! :+1:

PR #642, opened 2015-07-18
Fix bincount.jl - Float64 -> FloatingPoint

### LINKED ISSUE #641: Histogram plot doesn't work with float32 input arrays
This thing fails:
```julia
plot(x=float32(randn(100)), Geom.histogram())
```

````bincount_pll` has no method matching bincount_pll(::Int64, ::Int64, ::Array{Int64,1}, ::Float32)```
bincount_pll expects Float64 as input

maximsch2:
Fails on Float32 arrays overwise. Fixes #641

### DIFF
--- a/src/bincount.jl
+++ b/src/bincount.jl
@@ -29,7 +29,7 @@
 # Returns:
 #   Log-likelihood with Birge's penalty applied.
 #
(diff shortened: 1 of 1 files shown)


dcjones:
Thanks!

PR #652, opened 2015-08-04
WIP: Document rendering process

darwindarak:
This documents how a `plot` invocation gets turned into a Compose scene graph.  I'm hoping this will make it easier to refactor existing code, and encourage contributions to the code base (maybe even plugin packages).

The plan is to have a "10000-foot view" of the process, then a detailed function-by-function walkthrough.  I've only started on the overview section, and there are already many simplifications.  While many default are actually computed during the rendering process, the overview assumes that all the elements are already specified.

This is still a massive work in progress.  Any corrections/suggestions are welcome and appreciated.

@dcjones, do you mind checking for correctness?  Hopefully I haven't botched it too badly.

### DIFF
--- a/doc/dev_pipeline.md
+++ b/doc/dev_pipeline.md
@@ -1,0 +1,93 @@
+---
+title: Rendering Pipeline
+author: Darwin Darakananda
+part: Gadfly Internals
+order: 3001
+...
+
+```{.julia hide="true"}
+using DataFrames
+using Color
+using Compose
+using RDatasets
+using Showoff
+using Gadfly
+```
+
+How does the function call
+
+```{.julia execute="false"}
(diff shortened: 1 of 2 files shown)


dcjones:
I pushed a couple typo fixes, but it looks great. Really useful stuff.

timholy:
Any reason not to merge? Indeed looks useful!

darwindarak:
There are still quite a few sections I haven't had time to add.  If it is usable as it is now, maybe we can merge it first and use it as a base for further additions.

dcjones:
Sounds good. Feel free to add to this anytime.

dcjones:
Looks nice: http://gadflyjl.org/dev_pipeline.html :+1:

PR #653, opened 2015-08-05
Always set emit_on_finish to false in writemime

Keno:


dcjones:
I'm not sure I understand why `emit_on_finish` should be set to false.

shashi:
Is this related to https://github.com/dcjones/Compose.jl/pull/170 ?

Keno:
Yes, I believe so.

tlnagy:
I just confirmed that doesn't seem to fix https://github.com/GiovineItalia/Gadfly.jl/issues/747

PR #655, opened 2015-08-05
Allow x-axis with SIUnits

mbauman:
Since https://github.com/Keno/SIUnits.jl/pull/34, `one(10Second)` has simply been returning `1` instead of `1s`.  While this is technically more correct as a multiplicative identity, it caused issues in comparisons and non-multiplicative maths. This changes usage of `one(T)` in determining x bounds and ticks to `convert(T, one(T))`, which more explicitly expresses what we want - a one of the same type.

There are probably more cases like this, but it's a start.  The very simple `plot(x=(0:.1:10)Second, y=sin(0:.1:10)Meter)` now works!

### DIFF
--- a/src/ticks.jl
+++ b/src/ticks.jl
@@ -4,7 +4,7 @@
 # little opaque because I want to avoid assuming the log function is defined
 # over typeof(xspan)
 function bounding_order_of_magnitude{DT}(xspan::DT)
-    one_dt = one(DT)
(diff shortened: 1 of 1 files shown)


dcjones:
Thanks!

stevengj:
In 0.6, the right thing to do is to call `oneunit` if you want a dimensionful `one`.

PR #660, opened 2015-08-15
Typo in error handling code

joehuchette:
Without this, I was getting a really cryptic error message in ``sortperm!``.

### DIFF
--- a/src/aesthetics.jl
+++ b/src/aesthetics.jl
@@ -165,7 +165,7 @@ function assert_aesthetics_equal_length(who::String, aes::Aesthetics, vars::Symb
 
     n = length(getfield(aes, vars[1]))
     for i in 2:length(vars)
-        if length(getfield(aes, vars[1])) != n
+        if length(getfield(aes, vars[i])) != n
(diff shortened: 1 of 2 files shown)


joehuchette:
@dcjones this is failing the ``stat_qq.jl`` test with a call like
```jl
x = rand(Normal(), 100)
pl9 = plot(x=x, y=Normal(), color=rand(length(x)), Stat.qq, Geom.point)
```
When Gadfly is applying the ``QQStatistic``, it rewrites ``x`` and ``y`` to the output of ``Distributions.qqbuild``, which strips off two entries, leaving ``length(x) == length(y) == 98`` but ``color`` unchanged with ``length(color) == 100``. Any idea how to fix this?

joehuchette:
(Note that without this fix, ``assert_aesthetics_equal_length`` would never produce an error)

dcjones:
Thanks for fixing this. I'm thinking that I'll just disallow `color` with `Stat.qq`. I don't know of anything reasonable to do with it if `Stat.qq` is not 1-1.

joehuchette:
Seems reasonable. I looked into that, but couldn't figure out the right mechanism to disallow certain combinations of things like that.

PR #662, opened 2015-08-19
Avoid various depreciation messages

matthieugomez:


timholy:
Is this still relevant? Sorry there wasn't comment earlier.

PR #664, opened 2015-08-21
Switch to Colors (do not merge yet)

timholy:
This switches to the new Colors package, as discussed in JuliaLang/Color.jl#101. Posting for discussion. Do not merge until C-day, tentatively scheduled for Tuesday, Aug 25.

I do not expect this to pass tests yet, because of the inter-package dependency relationships. (EDIT: this passes tests locally for me on 0.4. Haven't tested 0.3 yet.) This is largely an automated translation, and there may be things that can be done better. I'll flag at least one of them below.

### DIFF
--- a/src/color_misc.jl
+++ b/src/color_misc.jl
@@ -12,7 +12,7 @@
 #   A weighted mean color of type T.
 #
 function weighted_color_mean{S <: Number}(
-        cs::AbstractArray{LAB{Float64},1}, ws::AbstractArray{S,1})
+        cs::AbstractArray{Lab{Float64},1}, ws::AbstractArray{S,1})
     l = 0.0
     a = 0.0
     b = 0.0
@@ -23,7 +23,7 @@ function weighted_color_mean{S <: Number}(
         a += w * c.a
         b += w * c.b
     end
-    LAB(l, a, b)
+    Lab(l, a, b)
 end
 
 
@@ -40,15 +40,15 @@ end
 #   n: number of colors
 #
 function lab_rainbow(l, c, h0, n)
-    ColorValue[LCHab(l, c, h0 + 360.0 * (i - 1) / n) for i in 1:n]
+    OpaqueColor[LCHab(l, c, h0 + 360.0 * (i - 1) / n) for i in 1:n]
 end
 
 function luv_rainbow(l, c, h0, n)
-    ColorValue[LCHuv(l, c, h0 + 360.0 * (i - 1) / n) for i in 1:n]
+    OpaqueColor[LCHuv(l, c, h0 + 360.0 * (i - 1) / n) for i in 1:n]
 end
 
 # Helpful for Experimenting
-function plot_color_scale{T <: ColorValue}(colors::Vector{T})
+function plot_color_scale{T <: OpaqueColor}(colors::Vector{T})
     println(colors)
     canvas(UnitBox(length(colors), 1)) <<
             (compose([rectangle(i-1, 0, 1, 1) << fill(c)
@@ -62,12 +62,12 @@ end
 # Generate a gradient between n >= 2, colors.
 
 # Then functions return functions suitable for ContinuousColorScales.
-function lab_gradient(cs::ColorValue...)
+function lab_gradient(cs::OpaqueColor...)
     if length(cs) < 2
         error("Two or more colors are needed for gradients")
     end
 
-    cs_lab = [convert(LAB, c) for c in cs]
+    cs_lab = [convert(Lab, c) for c in cs]
     n = length(cs_lab)
     function f(p::Float64)
         @assert 0.0 <= p <= 1.0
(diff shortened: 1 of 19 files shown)


src/color_misc.jl
```diff
@@ -40,15 +40,15 @@ end
 #   n: number of colors
 #
 function lab_rainbow(l, c, h0, n)
-    ColorValue[LCHab(l, c, h0 + 360.0 * (i - 1) / n) for i in 1:n]
+    OpaqueColor[LCHab(l, c, h0 + 360.0 * (i - 1) / n) for i in 1:n]
```
timholy: Is there an advantage in coercing these to a non-leaftype?

src/guide.jl
```diff
@@ -345,7 +345,7 @@ function render_continuous_color_key(colors::Dict,
          (context(),
           line([[(0, 1 - y), (swatch_width, 1 - y)] for y in values(colors)]),
           linewidth(theme.grid_line_width),
-          stroke(color("white"))),
+          stroke(parse(Color, "white"))),
```
timholy: This, and similar changes below, are not strictly necessary since I added `color` to Compose. I can drop this part if you prefer.

lobingera:
I can't comment on the code, but the (re-)implementation of Colors.jl wasn't driven by Gadfly afair. I'd recommend to work and PR on a less prominent plotting package first, make and regonize all possible errors and then move to the 'big' names like Gadfly and Images (e.g.).

timholy:
A major reason to do Gadfly first is that @dcjones is the original creator of Color.jl, and I'm particularly interested in making sure he's happy with the transition. (Comments, @dcjones? Let me know anything you're unhappy about.)

Second, Gadfly and Images are (AFAIK) the two most extensive users of colors, and I wanted to find out early what the problems will be rather than doing a large pile of easy-but-tedious work on other packages and then discovering issues and going back and re-doing that work. So I used Gadfly and Images as the stress tests.

This already passes all tests for me locally, so I think it's actually "done." Not that I don't expect to discover hidden problems (apologies in advance), but I think realistically this is the best that can be expected.

lobingera:
@timholy, you are right with testing for Gadfly and Images and i always forget that in git based development branching is easy (actually branching is easy in all vcs, but in git, merging is also easy and low cost...). A few lines up i see a travis failing and in general i'd like to see more: works messages from different environments (windows, OSX, div. Linux) first before publishing the change.

timholy:
Like I said at the top, it's going to fail all tests---even if the code is perfect---until C-day. This is a really difficult change to test. If Gadfly switches and Cairo doesn't, it will fail. If Cairo switches and Gadfly doesn't, it will fail. If we merge https://github.com/JuliaLang/Cairo.jl/pull/112 today and wait a week on this, then users won't be able to use the Cairo backend on Gadfly until then. The point is that we don't have a good way of testing large, coordinated changes across packages.

About the only way this can work is for people to look over the code, give their in-principle approval, and then I'll try to do sequential merges (running the Travis & AppVeyor tests) on C-day.

lobingera:
juliaLang/julia#12730

timholy:
Good idea to file that, thanks.

timholy:
Replaced by #667.

timholy:
@lobingera, see https://travis-ci.org/timholy/Dummy.jl/builds/77127046, which was run with
```
# Documentation: http://docs.travis-ci.com/user/languages/julia/
language: julia
sudo: required
os:
  - linux
julia:
  - release
  - nightly
notifications:
  email: false
# uncomment the following lines to override the default test script
script:
  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
  - julia -e 'Pkg.add("ColorTypes"); Pkg.add("Colors"); Pkg.add("ColorVectorSpace")'
  - julia -e 'Pkg.add("Tk")'
  - julia -e 'Pkg.add("Images"); Pkg.checkout("Images", "teh/colors")'
  - julia -e 'Pkg.add("Gadfly"); Pkg.checkout("Gadfly", "teh/colors2")'
  - julia -e 'Pkg.checkout("DataArrays")'
  - julia -e 'Pkg.checkout("Hexagons")'
  - julia -e 'Pkg.checkout("Compose", "teh/colors2")'
  - julia -e 'Pkg.add("Cairo"); Pkg.checkout("Cairo", "teh/colors2")'
  - julia -e 'Pkg.add("Graphics"); Pkg.checkout("Graphics", "teh/colors")'
  - julia -e 'Pkg.add("Winston"); Pkg.checkout("Winston", "teh/colors")'
  - julia -e 'Pkg.add("ProfileView"); Pkg.checkout("ProfileView", "teh/colors")'
  - julia -e 'Pkg.test("Images")'
  - julia -e 'Pkg.test("Cairo")'
  - julia -e 'Pkg.test("Graphics")'
  - julia -e 'Pkg.test("Gadfly")'
  - xvfb-run julia -e 'Pkg.test("Winston"); Pkg.test("ProfileView")'
```
Everything passed. Other packages (e.g., ColorBrewer, Escher, PyPlot) don't have Travis-testing enabled, so I think this is as good as we can hope for.

PR #667, opened 2015-08-24
Switch to Colors package (do not merge yet)

timholy:
Replacement for #664. Do not merge until C-day.

The last commit is mostly for fun, experimenting with whether we can use more concretely-typed containers. This represents what should be a path towards performance improvements for Gadfly (but I haven't tested). If done well, the only potential cost I'm aware of might be increased compilation times under some circumstances. (EDIT: meaning, it might have to compile multiple type-stable versions of some methods, whereas formerly it used only one type-unstable version.)

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -38,6 +38,12 @@ end
 
 typealias ColorOrNothing Union(Color, TransparentColor, Nothing)
 
+# Allow users to supply strings without deprecation warnings
+# Note this strips any transparency; is this desirable?
+parse_color(c::Colorant) = color(c)
+parse_color(str::AbstractString) = color(parse(Colorant, str))
+parse_color_vec(c...) = to_vec(map(parse_color, c)...)
+@noinline to_vec(c...) = [c...]
 
 element_aesthetics(::Any) = []
 default_scales(::Any) = []
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -4,7 +4,7 @@ VERSION >= v"0.4.0-dev+6641" && __precompile__()
 module Gadfly
 
 using Codecs
-using Color
+using Colors
 using Compat
 using Compose
 using DataArrays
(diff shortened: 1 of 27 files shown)


timholy:
Replaced by #667.

src/Gadfly.jl
```diff
@@ -36,8 +36,14 @@ function __init__()
 end
 
 
-typealias ColorOrNothing Union(ColorValue, AlphaColorValue, Nothing)
+typealias ColorOrNothing Union(Color, TransparentColor, Nothing)
 
+# Allow users to supply strings without deprecation warnings
+# Note this strips any transparency; is this desirable?
```
timholy: @dcjones, is this what you want here? Or is any `Colorant` OK?

timholy:
Again, the test failure on `release` is not anything to worry about (yet), because the tagged version of Cairo forces `Color` to load and that causes confusion with `Colors`. On `nightly`, that's a very odd precompilation failure, one I once saw on my own machine but which went away with a newer version of julia (so I just assumed it was a bug that had already been fixed). CCing @vtjnash and @stevengj.

stevengj:
@timholy, it looks like JuliaLang/julia#1250: `DataArrays` is not marked with `__precompile__()` and is apparently loaded (in non-precompiled form) by something in the test framework(?) before `Gadfly` and hence fails to load in its precompiled form when Gadfly is precompiled.

Solution: put `__precompile__()` in DataArrays.jl (assuming it is precompile-safe).

stevengj:
Some of the test files have `using RDatasets, DataArrays, Gadfly`.  Put `Gadfly` first and the problem should go away.

(But in general, whenever you are using a package X and are precompiling, it is best practice to get `__precompile__()` merged in package X.  Not only will that avoid this problem, but it also verifies with the upstream package author that X is indeed precompilable.

timholy:
Thanks for looking, @stevengj!

timholy:
@dcjones, if you need more time to review, I will push C-day back. Just let me know.

dcjones:
This looks good to me. Actually, it seems pretty safe to merge this at any time. It works fine with pre-transition Cairo.

timholy:
Thanks!

Combined testing here: https://travis-ci.org/timholy/Dummy.jl. Assuming this is on-track tomorrow, I'll merge.

protogeezer:
What is the difference between “pre-transition” Ciaro and what’s coming?

timholy:
@dcjones was pointing out that he was able to make it work even without switching Cairo to Colors. But both have been switched now.

PR #672, opened 2015-09-01
RFC: call plot and render once in module to precompile those routines

spencerlyon2:
Trying to speed things up.

On current master I get:

```
julia> tic(); using Gadfly; toc()
elapsed time: 3.452206278 seconds
3.452206278

julia> x = 1:10; y = cumsum(randn(10));

julia> gc_enable(false)
true

julia> @time render(plot(x=x,y=y,Geom.point));
 17.578433 seconds (20.63 M allocations: 973.109 MB)
```

With this PR:

```
julia> tic(); using Gadfly; toc()
elapsed time: 3.511360173 seconds
3.511360173

julia> x = 1:10; y = cumsum(randn(10));

julia> gc_enable(false)
true

julia> @time render(plot(x=x,y=y,Geom.point));
  7.876351 seconds (7.13 M allocations: 292.230 MB)
```

The disclaimer I offer here is that it takes much much longer to precompile Gadfly (something like 40 seconds instead of 20). For users this is a one time per update cost, so it would probably be worth it. For devs it is quite annoying to wait so long every time Gadfly is touched. We could get around that by just commenting out these lines while developing and enabling them again before updating master. Because that is more of a hassle for development I'm not sure that trade-off is worth it, hence the reason this is an RFC. 

If we do decide this is worthwhile we could probably add more `Geom`s or `Guide`s to what is `render`ed in the let block to get those things to be faster on the first call per session also.

Any thoughts?

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -1159,4 +1159,11 @@ end
 const x_axis_label_aesthetics = [:x, :xmin, :xmax]
 const y_axis_label_aesthetics = [:y, :ymin, :ymax]
 
+let
+    x = 1:10; y = cumsum(randn(10));
+    p = plot(x=x, y=y, Geom.point, Geom.line)
+    render(p)
+    nothing
+end
+
 end # module Gadfly


vchuravy:
Maybe instead Base.precompile would be helpful? But I see that is the approach #671 followed and I don't know why that was closed.

spencerlyon2:
Yeah [I tried](https://github.com/spencerlyon2/Gadfly.jl/tree/sl/breakups) something like that too. 

My strategy in the other branch was two fold:

- break large functions into small chunks
- Precompile those chunks as well all the `render` methods for `Geom`s and `Guide`s

I saw almost no speedup

I think the problem was that there are just so many functions those methods call out too and I didn't catch all of them in my `Base.precompile` net.

So, instead of spending a long time trying to figure out just what methods needed to be precompiled, I let the compiler do it for me! The takeaway I have is that we can trick the (pre)-compiler into making sure certain routines are precompiled by just running the code we want to be fast (in this case the call to `plot` and `render`).

dcjones:
Merged #673 which supersedes this, but thanks for investigating!

spencerlyon2:
Awesome, glad @timholy was willing to do by hand what I was too lazy to determine. I just made the compiler find the right functions to pre-compile, @timholy listed out all 700+ of them! :clap:

Issue #680, opened 2015-09-04
An intermediate level between Gadfly & Compose?


timholy:
I'm thinking about interactivity a fair bit these days. I'm still learning the code base, but I think I understand enough now to ask an architectural question: might we need a layer of "rendering" that lies between the top Gadfly level and Compose?

Here's one example: say I have a histogram with 10^9 points in it, and I want ~10^4 bins. I want to render this to some Gtk canvas that I can resize and zoom in on. If I have the canvas redraw starting from the top "Gadfly" level, it's going to have to re-bin the points, which from a performance perspective will be too awful to contemplate. So alternatively, I render to a `Compose.Context`, which makes redrawing very fast. However, now when I zoom in on a small region (my monitor does not have 10^4 pixels horizontally, so I can get more information by zooming), I can't update my tick labels because they've already been rendered in the `Context`.

I've played around with adding tags (`Geom.histogram(tag=:myhist)`) to `Form` objects and then searching the tree for tagged objects and updating them in-place. This is clearly a workable strategy. But I wonder if there's a general need for an intermediate "pre-processed" layer? As another example, when plotting a bunch of points in 2d, you might want to put them in a quad-tree so that updates get fast when you zoom in on a small region.

If there is an intermediate layer, it seems it could do things like computing data limits, etc. Basically any operation that will be of a different order than rendering itself could be.

dcjones:
I've thought a little about this. I think partial re-renders might get very complicated, but with the right indexed data structure, subsetting and binning operations could possibly be made cheap enough that the entire plot could be re-rendered efficiently enough. That could also allow plotting data that's too big to read entirely into memory. One option I've looked at is [sqlite's r-tree index](https://www.sqlite.org/rtree.html), but it might be more efficient to implement something from scratch.

On the gadfly side all that needs to be done is to generalize things like binning, minimum, and maximum which currently assume the input is an array, which seems fairly manageable.

timholy:
Looking at it a bit more, I think one fairly easy step in this direction is to split `render` into `render_prepare` followed by the stuff that happens starting at the call to `render_prepared`.

protogeezer:
I'd like to participate in the discussion by offering an alternative - I've met with Dan a couple of times to discuss the concept of replacing the Cairo based graphics layer with one based on Qt. Qt (vector) graphics would likely speed things up by making it possible to exploit the Qt notion of a graphics view, something completely missing from the Gnome/GTK window model. 

[...]

PR #683, opened 2015-09-08
Insert more entry points into render

timholy:
See #680. There is surely more work along those lines in the future, but this is a good start.

This also makes dynamic tick rendering optional, since this is the slowest part of guide-rendering and non-javascript renderers don't need it.

### DIFF
--- a/src/guide.jl
+++ b/src/guide.jl
@@ -45,6 +45,10 @@ end
 
 const background = PanelBackground
 
+function render(guide::Gadfly.GuideElement, theme::Gadfly.Theme,
+                aes::Gadfly.Aesthetics, dynamic::Bool=true)
+    render(guide, theme, aes)
+end
 
 function render(guide::PanelBackground, theme::Gadfly.Theme,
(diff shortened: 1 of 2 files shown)


src/guide.jl
```diff
@@ -47,7 +47,7 @@ const background = PanelBackground
 
 
 function render(guide::PanelBackground, theme::Gadfly.Theme,
-                aes::Gadfly.Aesthetics)
```
dcjones: Would it be easier to define a fallback function like this instead of adding so many unused parameters?
```julia
function render(guide::Guide, theme::Gadfly.Theme,
                aes::Gadfly.Aesthetics, dynamic::Bool=true)
    render(guide, theme, aes)
end
```

timholy: Sure, I can do that.

timholy:
Much cleaner this way, thanks for the suggestion!

dcjones:
Looks good, thank you!

PR #687, opened 2015-09-11
Compose.Image doesn't exists right now

maximsch2:
Latest version fails to load on 0.4 otherwise


timholy:
I'm guessing this is another one of those dependent-package (probably, Cairo) issues. @dcjones, just chew me out if I'm messing up by merging this.

timholy:
(And, thanks @maximsch2!)

maximsch2:
Thanks for merging it! It would be great to also tag a new version, as the latest one doesn't load now without Cairo installed.

dcjones:
Thanks Maxim and Tim. That was one of those conditional functions. I'm not sure how I missed it. I'm tagging a new version now.

PR #689, opened 2015-09-12
Preserve transparency in supplied colorants

timholy:
The matching element of the pair for https://github.com/dcjones/Compose.jl/pull/152


timholy:
(Tests will be expected to fail on this until that PR gets merged.)

dcjones:
Thanks!

Issue #392, opened 2014-08-11
Dashed Lines


dcjones:
Was #284, but GitHub wrongly thought I merged that.

tbreloff:
Just curious... has there been any progress on different line styles?

PR #695, opened 2015-09-18
Add line_style to Theme

tbreloff:
Addresses: #392 and probably others.

Theme will now accept a `line_style` argument.  If passed a vector, it converts into a sequence similar to [this example](https://github.com/dcjones/Compose.jl/blob/master/examples/dashedlines.jl).  If passed a symbol, it will return a pre-defined vector for that symbol (or throw an error if there isn't a pre-defined value for that symbol).

For line/hline/vline, if there is a value in the Theme for line_style, it will add a `strokedash` argument to the compose call.

Example:
```
using Plots
plot(cumsum(randn(10,5),1)+5, w=5, style=:auto, t=:line)
plot!(rand(1,10)*10, w=5, style=:auto, t=[:vline,:hline])
```

tmp


tbreloff:
Closing in favor of #702

Issue #614, opened 2015-06-23
Alpha blending


bdeonovic:
Is it possible to make colors transparent, i.e. alpha blending

bjarthur:
yes, use `RGBA`

Issue #152, opened 2013-12-31
PGFPlots backend


mykelk:
For publication quality plots, I really enjoy the [pgfplots](http://pgfplots.sourceforge.net/) package for latex.  It supports line plots, scatter plots, bar plots, area plots, mesh and surface plots, patch plots, contour plots, quiver plots, histogram plots, polar axes, etc. And all the text is rendered by latex. It might make a nice backend for Gadfly. Here is a very simple proof of concept (ideally, it would use the PDF directly, but I haven't gotten that to work yet).

``` julia
function plot(x, y)
    f = open("test.tex", "w")
    println(f,"\\documentclass{minimal}")
    println(f,"\\usepackage{pgfplots}") 
    println(f,"\\usepackage[active,tightpage]{preview}")
    println(f,"\\PreviewEnvironment{tikzpicture}")
    println(f,"\\usepackage[active,tightpage]{preview}")
    println(f,"\\begin{document}")
    println(f,"\\begin{tikzpicture}")
    println(f,"\\begin{axis}[scale=1.5,scale only axis]")
    println(f,"\\addplot coordinates {")
    for i = 1:length(x)
        println(f, "(", x[i], ",", y[i], ")")
    end
    println(f,"};")
    println(f,"\\end{axis}")
    println(f,"\\end{tikzpicture}")
    println(f,"\\end{document}")
    close(f)
    success(`pdflatex test`);
    success(`convert -density 300 test.pdf -resize 25% test.png`)
    imread("test.png")
end
```

Here is what it looks like in an IJulia notebook.

![image](https://f.cloud.github.com/assets/6294079/1827828/7aa05b7c-7242-11e3-9470-2fe690b1b01c.png)

bjarthur:
closing at it seems resolved

PR #698, opened 2015-09-19
bar_opacity Theme property and support in Geom.bar to render blended bars

penntaylor:
Simple change that allows plotting of overlapping histograms.

Example use:
```julia
using Gadfly
using Distributions

aa = rand(Normal(), 3000)
ab = rand(Normal(1.5, 1), 3000)

plot(layer(x=aa, Geom.histogram,Theme(default_color=colorant"#ffccee",bar_opacity=0.7)),
     layer(x=ab, Geom.histogram,Theme(bar_opacity=0.7)))
```
which produces:
bar_opacity
Minor issue: Drawing to SVG target works correctly, but PNG target isn't quite right. (I made no changes to any of the draw targets; does this behavior show up with ribbons as well?) For the above image, I saved as svg and afterwards converted to png.

### DIFF
--- a/src/geom/bar.jl
+++ b/src/geom/bar.jl
@@ -81,7 +81,7 @@ function render_colorless_bar(geom::BarGeometry,
             svgclass("geometry"))
     end
 
-    compose!(ctx, fill(theme.default_color))
+    compose!(ctx, fill(theme.default_color), fillopacity(theme.bar_opacity))
     if isa(theme.bar_highlight, Function)
         compose!(ctx, stroke(theme.bar_highlight(theme.default_color)))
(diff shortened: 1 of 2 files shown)


penntaylor:
Related to issue #614: Alpha blending

penntaylor:
Nevermind. I see now that alpha values on `default_color` actually work for this in master, they just didn't work yet in release 0.3.16. My apologies for the excess noise here.

tbreloff:
No worries. I just did the same thing. I spent a couple hours trying to figure out why alpha values weren't propagating properly and then I found  PR #152, checked out master, and it worked.

PR #702, opened 2015-09-21
Add line_style to Theme

tbreloff:
Note: this includes the recent changes to fix deprecated bindings.

Addresses: #392 and probably others.

Theme will now accept a line_style argument. If passed a vector, it converts into a sequence similar to this example. If passed a symbol, it will return a pre-defined vector for that symbol (or throw an error if there isn't a pre-defined value for that symbol).

For line/hline/vline, if there is a value in the Theme for line_style, it will add a strokedash argument to the compose call.

Example:
```
using Plots
plot(cumsum(randn(10,5),1)+5, w=8, style=:auto)
plot!(rand(1,10)*10, w=8, style=:auto, t=[:vline,:hline])
```
tmp

### DIFF
--- a/src/geom/hline.jl
+++ b/src/geom/hline.jl
@@ -27,10 +27,13 @@ function render(geom::HLineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics
     color = geom.color === nothing ? theme.default_color : geom.color
     size = geom.size === nothing ? theme.line_width : geom.size
 
+    line_style = theme.line_style == nothing ? [] : [strokedash(Gadfly.getStrokeVector(theme.line_style))]
+
     return compose!(
         context(),
         Compose.line([[(0w, y), (1w, y)] for y in aes.yintercept], geom.tag),
         stroke(color),
         linewidth(size),
-        svgclass("xfixed"))
+        svgclass("xfixed"),
+        line_style...)
 end
--- a/src/theme.jl
+++ b/src/theme.jl
@@ -62,6 +62,20 @@ function default_middle_color(fill_color::TransparentColor)
         fill_color.alpha)
 end
 
+getStrokeVector(vec::AbstractVector) = vec
+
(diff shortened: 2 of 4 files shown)


src/geom/hline.jl
```diff
@@ -27,10 +27,13 @@ function render(geom::HLineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics
     color = geom.color === nothing ? theme.default_color : geom.color
     size = geom.size === nothing ? theme.line_width : geom.size
 
+    line_style = theme.line_style == nothing ? [] : [strokedash(Gadfly.getStrokeVector(theme.line_style))]
```
dcjones: `line_style` doesn't need to wrapped be a vector. Calling `compose` on `nothing` is a nop.

src/theme.jl
```diff
@@ -62,6 +62,20 @@ function default_middle_color(fill_color::TransparentColor)
         fill_color.alpha)
 end
 
+getStrokeVector(vec::AbstractVector) = vec
```
dcjones: Could you rename to `getstrokevector` or `get_stroke_vector` for consistency?

dcjones:
Thanks for doing this. A couple nit picks noted.

chkwon:
@tbreloff How should I use the `line_style` argument? I've tried the following, which didn't work:

    using Gadfly
    plot(x=rand(10), y=rand(10), Geom.line, Theme(line_style=:dash))

chkwon:
I found this works.

    plot(x=rand(10), y=rand(10), Geom.line, Theme(line_style=Gadfly.get_stroke_vector(:dashdot)))

I'll just leave my comments, for any one who will need help.

tbreloff:
Currently this is the best way. I only added the bare minimum to this PR to implement functionality I wanted through Plots.jl (which I recommend ;)

PR #703, opened 2015-09-22
Reset pan and zoom on double click

Godisemo:
This adds the functionality to reset the zoom and pan of a plot by double clicking, like it works in Matlab.


dcjones:
Looks good, thank you.

PR #704, opened 2015-09-24
Union() -> Union{}

timholy:
Companion of https://github.com/dcjones/Compose.jl/pull/158

There are still a few concatenation and
```
WARNING: Base.Uint8 is deprecated, use UInt8 instead.
```
warnings (presumably due to dependencies), but this gets rid of the vast majority of them.


dcjones:
Thanks!

PR #707, opened 2015-09-25
Fix bouncing in Safari when zoom scrolling

Godisemo:
This fixes the bouncing effect that is present in Safari (possibly other browsers).


dcjones:
Cool, thanks for fixing my shoddy javascript. :smile:

Godisemo:
Thanks. This is part of a project of mine to make the interactive plots more usable. Expect more to come in the near future!

StefanKarpinski:
> Expect more to come in the near future!

:tada:

timholy:
@Godisemo, see also [Immerse.jl](https://github.com/JuliaGraphics/Immerse.jl), which has quite a few interactive features that are otherwise lacking (videos posted https://www.youtube.com/channel/UCFkqQf_Ie5PN4l146ajnujg). I've been thinking about writing a browser-based backend for it, but am a newbie when it comes to browsers.

In my opinion, the really important thing is to be able to write the callbacks in Julia rather than javascript---otherwise we're back to a two-language problem, which is bad because it's what Julia was created to avoid in the first place.

If you'd like to collaborate on a browser-based backend for Immerse, that would be awesome.

Godisemo:
@timholy I did not know about Immerse.jl until a few days ago when I stumbled upon it. I have not tested it yet since I didn't manage to install the Gtk.jl package on my setup. I watched your videos though. From what I understand the power of immerse is that it makes interacting with plot data possible. All other interactions such as zooming and panning that doesn't interact with the data should in my opinion be included in Gadfly itself. This is useful for example when you want to export a plot and use it on a website and still keep it interactive. I am definitely open for discussion regarding this.

timholy:
OSX users had a broken Gtk, but I think that got fixed yesterday.

If panning and zooming is all you want, sure, it's reasonable to have that in Gadfly. Immerse's main added value is interacting with the canvas.

PR #708, opened 2015-09-26
Add box zoom functionality

Godisemo:
This adds a first version of box style zooming, as shown in the images below. It also works as expected when one or both dimensions are non-scalable. To initialize the zoom start dragging with the shift key pressed. You can cancel the zoom action by pressing escape (also works for panning).

As of now the box zoom is constrained to the the aspect ratio of the plot. I plan to fix this later by separating the scale factor for x and y.

skarmavbild 2015-09-26 kl 12 51 21
skarmavbild 2015-09-26 kl 12 51 29
skarmavbild 2015-09-26 kl 12 51 30

### DIFF
--- a/src/gadfly.js
+++ b/src/gadfly.js
@@ -138,42 +138,10 @@ Snap.plugin(function (Snap, Element, Paper, global) {
             .drag(Gadfly.guide_background_drag_onmove,
                   Gadfly.guide_background_drag_onstart,
                   Gadfly.guide_background_drag_onend);
-        var plot = this;
-        this.node.addEventListener("statechanged", statechanged);
-        window.addEventListener("keydown", function(e) { keyfunction(plot, e); });
-        window.addEventListener("keyup", function(e) { keyfunction(plot, e); });
         return this;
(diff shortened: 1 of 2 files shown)


dcjones:
Thanks, this is lovely.

My only concern is that calling `init_pan_zoom` at load time can cause a page with many plots to hang for a long time loading. For an example, run the tests then open `test/output/test.html`.

Godisemo:
Yeah, thats an issue I did not consider. Wouldn't it be enough to call `init_pan_zoom` on mouse enter though? All interactions require that the mouse is in the plot. I updated the PR with a commit that calls `init_pan_zoom` lazy on mouse enter. Also fixed a few javascript related bugs I found as well as one that I introduced.

Almost forgot, I also added so that you can zoom with + and - when hovering a plot.

Godisemo:
For future commits, would it have been better to split this PR into several smaller PRs?

dcjones:
Looks good. It's fine to do PRs this size, but smaller is also ok.

PR #713, opened 2015-10-03
Change mapping between scale and slider position

Godisemo:
This PR changes the function that maps scale to slider position and vice versa in an attempt to make the code easier to understand.
The new function is smooth over the interval, compared to the old function that was defined piecewise.

I don't know if there was some thought behind this function earlier, but I think that the break is a bit strange.

The new function is shown in blue and old function in yellow.
hej


dcjones:
Yes, this is definitely better. I don't think there was a particularly good reason for the old behavior.

PR #734, opened 2015-10-24
Fix deprecated syntax and use of Pkg.dir in the tests

yuyichao:
(Also some white space)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,16 @@
-language: cpp
-compiler:
-    - clang
+language: julia
+os:
+  - linux
+  - osx
+julia:
+  - 0.4
+  - nightly
 notifications:
-    email: false
-env:
-    - JULIAVERSION="juliareleases"
-    - JULIAVERSION="julianightlies"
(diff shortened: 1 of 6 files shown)


yuyichao:
The only remaining depwarn in the test run is https://github.com/JuliaStats/Distributions.jl/pull/431

yuyichao:
I updated the travis settings to use `language: julia` which should be a much more relable/newer version. I also disabled `Compose` checkout because `he really shouldnt be checking out master of compose if he wants tests to be release representative` -- Tony. Is there a good reason the test is running against Compose master (e.g. are you always releasing the two at the same time)?

yuyichao:
Note that the nightly test is expected to fail but it should be fixed by https://github.com/dcjones/Compose.jl/pull/164 .

yuyichao:
I put the checkout back since it looks like the two are meant to be in sync.

dcjones:
Thanks! They are meant to be in sync, I basically always tag them simultaneously.

PR #746, opened 2015-11-18
Fixed y-axis bug for histogram2d

kshyatt:
The size of the boxes for the `Geom.histogram2d` aesthetic was totally wrong - it was using the spacing for the x direction!


yuyichao:
Travis nightly should be fixed by https://github.com/dcjones/Loess.jl/pull/12

shashi:
:+1:

PR #757, opened 2015-12-01
Use parse_colorant to handle arbitrary inputs (fixes #756)

### LINKED ISSUE #756: Stack overflow due to new interaction with Colors
Reproduce with

```julia
plot(layer(xintercept=[4,5,6], Geom.vline(color=colorant"gray50")))
```

```
ERROR: StackOverflowError:
 in parse at /Users/idunning/.julia/v0.4/Colors/src/parse.jl:136 (repeats 80000 times)
```

This used to work, now it does not. Is that expected?

@timholy, any ideas?

timholy:
CC @IainNZ. This is a more Gadfly-specific fix than https://github.com/JuliaGraphics/Colors.jl/pull/239, but it's unclear to me which should be preferred.

`parse_colorant` is [defined in Compose](https://github.com/dcjones/Compose.jl/blob/55eabfe21ed94c1a8b88babd4abfb444fb695ef6/src/Compose.jl#L64-L66), and is intended to provide the fallback behavior without introducing "dirty" methods to the Base-defined `parse` function.

### DIFF
--- a/src/geom/hline.jl
+++ b/src/geom/hline.jl
@@ -7,7 +7,7 @@ immutable HLineGeometry <: Gadfly.GeometryElement
     function HLineGeometry(; color=nothing,
                            size::@compat(Union{Measure, (@compat Void)})=nothing,
                            tag::Symbol=empty_tag)
-        new(color === nothing ? nothing : parse(Colorant, color),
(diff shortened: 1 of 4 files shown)


timholy:
Superseded by https://github.com/JuliaGraphics/Colors.jl/pull/233

PR #777, opened 2015-12-29
Make spy() output nicer plots by default

mortenpi:
A small rewrite of the `spy()` function to get better defaults. The differences compared to the current one are
  - the axes now are now compactly around the matrix, whereas previously there
    was quite a lot of empty space, depending on the dimensions of the matrix
  - the aspect ratio of the axes is 1:1 by default, since squares look better
  - zeroes now get drawn, but NaNs get filtered out (currently zeros leave holes and NaNs make it crash)

Everything can still be overridden by passing a custom `Coord` or `Scale`, so I do not think that this forces the user into anything, but these are better defaults if you just want to take a quick look at some arbitrary matrix.

I.e. we go from this
spytest
to this
spytest-fixed
Plots produced by this simple script: https://gist.github.com/mortenpi/24730ed5f1b6de8976a4

### DIFF
--- a/src/poetry.jl
+++ b/src/poetry.jl
@@ -119,24 +119,67 @@ end
 
 # Simple heatmap plots of matrices.
 #
+# It is a wrapper around the `plot()` function using the `rectbin` geometry.
+# It also applies a sane set of defaults to make sure that the plots look nice
+# by default. Specifically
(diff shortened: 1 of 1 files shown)


darwindarak:
Looks great!  I'll merge it in a day or two if no objections are raised.

simleb:
:+1:

PR #779, opened 2015-12-30
Doc fixes and removal of ContinuousColorScale

simleb:
I implemented the changes around `ContinuousColorScale` and `color_continuous` discussed in #763. I took the opportunity to fix a bunch of typos and update a few things in the docs as well.

I don't know how but the docs will need to be regenerated at some point. I suppose only @dcjones can do that.

I chose `colormap` as the new keyword for the function that maps values in [0,1] to a color (my Matlab background is showing here, sorry). The name is of course open to discussion.


timholy:
Wow, that's a lot of doc fixes. Thanks!

Issue #786, opened 2016-01-13
Coloring fails if variable names contains greek symbols


montyvesselinov:
Plotting this data frame fails when there is alpha in the parameter column using this command:

Gadfly.plot(vdf, x="x", y="y", Gadfly.Geom.line, color="parameter")

| Row | x | y | parameter |
| --- | --- | --- | --- |
| 1 | 0.0 | 0.0 | "α_1" |
| 2 | 9.2199 | 581.326 | "α_1" |
| 3 | 19.1489 | 587.034 | "α_1" |
| 21 | 0.0 | 0.0 | "a2" |
| 22 | 9.2199 | 577.306 | "a2" |
| 23 | 19.1489 | 584.726 | "a2" |

montyvesselinov:
These are the errors I am getting:

ERROR: LoadError: LoadError: ArgumentError: invalid ASCII sequence
 in convert at /Applications/Julia-0.4.2.app/Contents/Resources/julia/lib/julia/sys.dylib
 in call at /Users/ftuser/.julia/v0.4/Compose/src/property.jl:476
 in svgclass at /Users/ftuser/.julia/v0.4/Compose/src/property.jl:488
 in render at /Users/ftuser/.julia/v0.4/Gadfly/src/geom/line.jl:205
 in render at /Users/ftuser/.julia/v0.4/Gadfly/src/geometry.jl:48
 in render_prepared at /Users/ftuser/.julia/v0.4/Gadfly/src/Gadfly.jl:803
 in render at /Users/ftuser/.julia/v0.4/Gadfly/src/Gadfly.jl:742
 in draw at /Users/ftuser/.julia/v0.4/Gadfly/src/Gadfly.jl:842

montyvesselinov:
The problem is parameter column is expected to be ASCIIString. However, in this case is AbstractString, which includes UTF8String.

bjarthur:
this now works for me

```
plot(x=[1,2,3,4],y=[1,2,3,4],color=["α_1","α_2","α_1","α_2"],Geom.point)
```

please reopen this issue if you still have problems.

PR #788, opened 2016-01-13
RFC: Remove ASCII requirement on strings when possible

darwindarak:
As shown in #786, using `ASCIIString` might be too restrictive.  While the current change will allow the specific test case in #786 to run, we might want to consider replacing all the `ASCIIString`'s with `AbstractString`'s.  Can anyone think of a reason why that might not be a good idea?


timholy:
Sorry I didn't notice this earlier. There's a lot to like about this change, with the one negative being type-stability, which affects performance: `AbstractString` is, well, abstract, and hence the compiler can't optimize code.

I guess the question is: could we make it `UTF8String`, or do we also want to support things like `SubString`, etc? If this is really just about unicode, then converting everything to `UTF8String` seems like it would maintain type-stability yet still support unicode.

tlnagy:
I don't think this solution will work anymore

Issue #328, opened 2014-06-19
Provide a way to draw tick marks


felixjung:
I like the ggplot approach taken by Gadfly. However, I dearly miss the ability to draw tick marks on my axes. For my publication purposes, I am required to not show grid lines and instead draw tick marks on the axes. Is there a specific reason they are missing at the moment?

rtambone:
Hi all, I just started working with Julia. I would need ticks marks without grid lines and up to date it seems this was not implemented. Any of you can help me?

PR #792, opened 2016-01-14
Add tick marks to plots

andrewdsen:
Provides functionality for tick marks (issue #328).
* Ticks are toggled with the theme keyword `tick_color`.
* Ticks are drawn over grid lines and have the same line width as grid lines.
* Tick length is set to 1% of the maximum axis length. (It would be easy to allow user definition via theme if that functionality is desired).

These look best when `panel_stroke` is defined. On that note, I also added a line width for the panel stroke (set to `theme.line_width`). This doesn't seem to affect things in the SVG but it makes a big difference in appearance in PDF rendering, where the panel line width is somewhat chunky.

I should note that the dynamic behavior in the SVG output is somewhat broken, in my opinion: ticks move when the window is panned but only the original ticks are retained (none appear to be generated when zooming). Not sure what to change to remedy this.


bjarthur:
@andrewdsen can you please rebase this PR with master?  i'd like to try out tick marks.  thanks.

Mattriks:
also now #1047.  Perhaps time to create a new branch, and modify/update the code in this PR.

PR #796, opened 2016-01-22
Update doc template to work with live Judo

amellnik:
This is the most recent template from Judo, also remove including Color
which is in base on 0.4.


tlnagy:
We're moving away from Judo for documentation, see progress at #879

Issue #610, opened 2015-06-20
Ternary plot


simleb:
I would love to be able to make [ternary plots][1] in Gadfly.

[1]: https://en.wikipedia.org/wiki/Ternary_plot

It is supported by a different kind of axes so where does this fit in the “grammar of graphics”? Should it be a Scale? A Geometry?

simleb:
Ideally, it would support multiple geometries. Point, Path, but also histogram2d would be nice, maybe coupled with tribin (analogous to rectbin but triangular).

dcjones:
I would like to add this. It does fit into a gog approach, in fact, Wilkinson's book has a section (9.2.3) on ternary plots.

I'd have to think a little about the implementation, but probably there'd be a `Coord.triangular` element, and new aesthetics named to avoid overlapping cartesian aesthetics, probably `x1, x2, x3`. Then it would be sufficient to say `plot(x1=a, x2=b, x3=c, Geom.point)`.

There would be a fair amount of work, because one would have to implement guides to draw the grid lines, write separate render functions for applicable geometries, etc. I would probably rearrange things so that render functions for geometries and guides dispatch on the coordinate type of the plot to avoid mixing of coordinates.

FJValverde:
Hi. Could one of the main developers of this enhancement provide an update about this? I regularly us ternary diagrams for a type of assessment plot in Machine Learning and would like to use this in Julia. I have developed such a thing for Matlab but in R-ggplot2 had to used somebody's implementation of the ternary geometry (http://www.ggtern.com).

PR #799, opened 2016-01-30
WIP: Polar coordinates

simleb:
I'm trying to implement polar coordinates. I followed @dcjones's suggestion in #610 (although it was for ternary coordinates) to add a new Coord and to make geometries and guides dispatch on the coordinate type. So far this is just a proof of concept, help would be more than welcome!

I introduced two new aesthetics: `rho` (aliases: `r` and `ρ`) and `phi` (alias `ϕ`), so one doesn't have to specify `Coord.polar` explicitly. I followed [ISO 31-11][ISO] for the names.

[ISO]: https://en.wikipedia.org/wiki/ISO_31-11

I chose to dispatch `element_coordinate_type` on the mapped aesthetics as well as the geometry. Not sure if this is the best way.

The biggest issue is with Guides. I don't understand the scale at which they are drawn. Right now the circular grid is too small while the angular grid doesn't even show up.

Eventually, better Stats need to be implemented to compute the grid correctly but I would like to have something basic working.

Here is what I have so far:

```julia
plot(ρ=linspace(0, 5, 50), ϕ=linspace(0, 5, 50), Geom.point)
```
polar


tlnagy:
Hey @simleb, are you still working on this? This would be a great thing to have in Gadfly.

simleb:
No, I'm afraid I won't have time to work on it for a while. Feel free to take over.

PR #802, opened 2016-02-05
Temporarily change the manual link and rm missing image

amellnik:
This is a stopgap until we can update gadflyjl.org to include the most recent docs.


darwindarak:
Thanks!

PR #805, opened 2016-02-11
Added basic inline help to exported methods based on existing inline/online documentation

mdpradeep:


darwindarak:
This is great, thanks!  I really like the known bugs section.

Issue #791, opened 2016-01-14
geom.vector?


gaechter:
i think up to now its not possible to draw a vector in Gadfly. It were nice to have in geometry an option geom.vector. 
Thank You for your great job!

Albert

Mattriks:
Here is a quick implementation of vectors/arrows/segments. For this initial implementation, I've simply tried to make use of the aesthetic args that already exist.  Suggestions for improvements welcome!  I think the following example shows how to use it.  
You can turn off the arrows by removing the `shape=true` argument from a layer.  Line_style is controlled through Theme (and is not currently an aesthetic).

You'll need this file: 
[Geom_segments.txt](https://github.com/dcjones/Gadfly.jl/files/206223/Geom_segments.txt)
Rename it: Geom_segments.jl


And here is an example:

```julia
using DataFrames
include("Geom_segments.jl")
function expand_grid(n) 
    q = linspace(-10, 10, n)
    return vcat([[i j] for i in q, j in q]...)
end

function dsinc(x,y, sc) 
    z = hypot(x,y)
    dz = cos(z)./z  - sin(z)./(z.^2)
    θ = atan2(y,x)
    dx = sc*abs(dz).*cos(θ)
    dy = sc*abs(dz).*sin(θ)
    return DataFrame(xa=x-dx, ya=y-dy, xb=x+dx, yb=y+dy, dz=dz)
end

function sincf(x, y) 
    z = hypot(x,y)
    z = sin(z)./(z)
    return DataFrame(x=x, y=y, z=(z))
end


X = expand_grid(50)
X2 = expand_grid(20)

D1 = sincf(X[:,1], X[:,2])
D2 = dsinc(X2[:,1], X2[:,2], 2.0)

theme1 = Gadfly.Theme(default_color=colorant"black", line_style=[1mm, 0.0mm])
theme2 = Gadfly.Theme(line_style = [0.75mm, 0.5mm])

coord = Coord.cartesian(xmin=-10.0, xmax=10.0, ymin=-10.0, ymax=10.0)
l1 = layer(D1, x=:x, y=:y, color=:z, Geom.rectbin, order=1)
l2 = layer(D2, x=:xa, xmax=:xb, y=:ya, ymax=:yb, shape=true,
            Geom_segment, theme1, order=2 )
l3 = layer(D2, x=:xa, xmax=:xb, y=:ya, ymax=:yb, color=:dz, shape=true,
        Geom_segment, theme1, order=2)


[...]

Issue #608, opened 2015-06-17
Vector field


ataias:
Firstly, I am not sure where to ask questions about Gadfly, that is why I am posting an issue here. Well, I was just wondering if there is a way to plot 2D vector fields in Gadfly.

dcjones:
There's not currently, though this is certainly something that would be worth adding.

bjarthur:
see https://github.com/GiovineItalia/Gadfly.jl/pull/992

PR #823, opened 2016-04-11
Segment Geometry

Mattriks:
A geometry for line segments/vectors/arrows. There is an example in the file geom_segment.md. This feature was previously requested in Issues #608 and #791. My current code will work for regular plots, but not yet for subplot_grid.  I'll attempt to implement that sometime soon!


tlnagy:
Hey @Mattriks, if you have time, can you rebase this on the latest master? Lets check if it works with all the recent changes and make sure it's all okay on 0.5. I think this would be a cool geometry to have.

Mattriks:
I've tested this now  on Julia 0.5-rc2. The example in the file geom_segment.md works.  Also, this Segment Geometry works when using with `Geom.subplot_grid`.

lobingera:
Just note: When i click on the issue, i first get

> 
> We went looking everywhere, but couldn’t find those commits.
> 
> Sometimes commits can disappear after a force-push. Head back to the latest changes here.

tlnagy:
@Mattriks Awesome. Could you add some tests?

tlnagy:
I got to play around with this and looks good! 

The only thing is that the arrowheads when calling `Geom.vector` look a little funky:

```julia
plot(x=rand(1:10, 10), y=rand(1:10, 10), xend=rand(1:10, 10), yend=rand(1:10, 10), Geom.vector)
```

screen shot 2016-08-24 at 12 26 59

And they don't behave properly when one of the axes is logged

screen shot 2016-08-24 at 12 30 13

tlnagy:
Actually, the first one is caused by not having a square axis ratio. But in general, I would think that you don't want scaling and transformations to apply to the arrowheads.

Mattriks:
I've implemented a new and improved `Geom.vector` (I've also included tests). In Gadfly, the current situation is that no information about `Scales` is passed to the geometry [render](https://github.com/dcjones/Gadfly.jl/blob/f6183acb6015c8e133e335d32227f551056c59f8/src/geometry.jl#L48) functions. But for rendering vectors properly, info about the `x` and `y` scales must be provided. That is, the user must manually provide a `Scale` object (with a `minvalue` and `maxvalue`) for both axes. I will look into automating this in the future. Here is a current working example:
```julia
srand(1234)
D = convert(DataFrame, 99*rand(4, 4)+0.5)
xsc  = Scale.x_continuous(minvalue=0.0, maxvalue=100)
xlogsc  = Scale.x_log10(minvalue=1, maxvalue=100)
ysc  = Scale.y_continuous(minvalue=0, maxvalue=100)
xsqrtsc  = Scale.x_sqrt(minvalue=0, maxvalue=100)

pa = plot(D, x=:x1, y=:x2, xend=:x3, yend=:x4, Geom.vector, xsc, ysc,
Guide.title("Plain scales"), Guide.ylabel(nothing), Guide.xlabel(" "))
pb = plot(D, x=:x1, y=:x2, xend=:x3, yend=:x4, Geom.vector, xlogsc, ysc,
Guide.title("Semi-log"), Guide.ylabel(nothing), Guide.xlabel("x"))
pc = plot(D, x=:x1, y=:x2, xend=:x3, yend=:x4, Geom.vector, xsqrtsc, ysc,
Guide.title("√x"), Guide.ylabel(nothing), Guide.xlabel(" "))
```
gvec

tlnagy:
Very nice. What do you think about filled-in arrowheads? Whenever I think of vector fields, I always think of something like this:
2000px-vectorfield svg

(https://en.wikipedia.org/wiki/Vector_field)

Mattriks:
Some extra attributes (example below). Wrt gradient vectorfields, that could be implemented as function `Geom.gradvectorfield(z, x, y)` (returning a `SegmentGeometry`), just as `Geom.contour(z,x,y)` returns a `LineGeometry`. That's a feature for the future, need to get `Geom.segment` merged first.
```julia
srand(123)
D = convert(DataFrame, 99*rand(4, 4)+0.5)
D[:col] = ["a","b","c","d"]

xsc  = Scale.x_continuous(minvalue=0.0, maxvalue=100)
ysc  = Scale.y_continuous(minvalue=0.0, maxvalue=100)
theme1 = Theme(line_width=3pt, line_style=[4mm,1mm], key_position=:none)
lf(x) = layer(x=:x1, y=:x2, xend=:x3, yend=:x4, x, color=:col)

pa = plot(D, lf(Geom.vector), xsc, ysc, Theme(key_position=:none))
pb = plot(D, lf(Geom.vector(filled=true)), xsc, ysc, theme1)
```      
vec2

tlnagy:
That looks awesome. Personally, I like filled arrows to be the default, but that can changed in the future.

I've looked through this and I think it all looks good. LGTM. @shashi what do you think?

src/geom/segment.jl
```diff
@@ -0,0 +1,117 @@
...
+#    line_style = Gadfly.get_stroke_vector(theme.line_style)
+    line_style = theme.line_style 
+    if is(line_style, nothing)
+        line_style = [] 
+    end
+
+
+# Geom.vector requires information about scales
+
+if geom.arrow
```
tlnagy: Is the top-level if statement necessary because `render` doesn't get scale information right now? Because this looks a bit hacky.

Mattriks:
`Scale`s are needed to stop arrowheads looking strange. Segments without arrowheads are not affected  i.e.  `Geom.segment(arrow=false)`

tlnagy:
I see. I guess eventually we'll want to get rid of that when we pass Scales to `render`.

tlnagy:
Alright. LGTM.

PR #838, opened 2016-05-05
Fix ArgumentError when plotting Geom.bar with empty array of values

JohnHBrock:
The below code was resulting in `ArgumentError: reducing over an empty collection is not allowed` because there is no value where category1 is B and category2 is Y.

```julia
using Gadfly

Gadfly.plot(x=[1, 4, 2],
            y=[5, 7, 3],
            xgroup=["A", "A", "B"],
            ygroup=["X", "Y", "Y"],
            Geom.subplot_grid(Geom.bar))
```

I fixed this by copying the logic used in the HistogramStatistic version of apply_statistic and added a test.

### DIFF
--- a/src/statistics.jl
+++ b/src/statistics.jl
@@ -196,6 +196,14 @@ function apply_statistic(stat::BarStatistic,
     end
 
     values = getfield(aes, var)
+    if isempty(getfield(aes, var))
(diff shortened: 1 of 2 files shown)


tlnagy:
Thanks for this fix, I'm rebooting travis and I'll merge this once tests pass. Sorry this took so long!

PR #877, opened 2016-08-22
saner title spacing in SVG, fixes #863

### LINKED ISSUE #863: Titles will spill over the diagram with newlines in `Guide.title()`
When newlines are used in `Guide.title()`, the title will overlay the diagram.

I seem to have problems uploading a screenshot, but try running this code:

```Julia
using Gadfly

plot(x=1:100,y=1:100, Guide.title("First Line\nSecond Line\nThird Line\n"))
```

tlnagy:
Currently, multiline titles simply spill-over onto the plot, which seems like a poor general solution. I believe a saner behavior is to increase the padding for the title above the plot:

New behavior:
screen shot 2016-08-22 at 11 55 19

Old behavior:

screen shot 2016-08-22 at 11 56 42

We might also want to consider exposing the padding variable to `Theme` so this becomes user adjustable.

### DIFF
--- a/src/guide.jl
+++ b/src/guide.jl
@@ -1045,8 +1045,8 @@ function render(guide::Title, theme::Gadfly.Theme,
 
     padding = 2mm
     ctx = compose!(
-        context(minwidth=text_width, minheight=text_height + 2padding),
-        text(0.5w, 1h - padding, guide.label, hcenter, vbottom),
+        context(minwidth=text_width, minheight=text_height + padding),
+        text(0.5w, 1h - text_height - padding, guide.label, hcenter, vtop),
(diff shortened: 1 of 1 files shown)


shashi:
Moving padding to `Theme` is a good idea.

src/guide.jl
```diff
@@ -1045,8 +1045,8 @@ function render(guide::Title, theme::Gadfly.Theme,
 
     padding = 2mm
     ctx = compose!(
-        context(minwidth=text_width, minheight=text_height + 2padding),
```
shashi: I personally like this change. The extra padding on top before this merge didn't feel quite right.

PR #879, opened 2016-08-23
WIP: transitioning docs to Documenter.jl

tlnagy:
Working on #878. This fixes #852, #865, and probably many others.

screen shot 2016-08-22 at 23 10 40

Couple things left

- [x] Fix regressions in `src/lib/guide_manual_color_key.md`
- [x] in `src/man/plotting.md`
- [x] Get plotting working https://github.com/JuliaDocs/Documenter.jl/issues/229
- [x] Set up Travis to auto-generate documentation for stable and release versions of Gadfly
- [ ] Too many items in navbar https://github.com/JuliaDocs/Documenter.jl/issues/230
- [x] Add Author information back in
- [x] Update README
- [x] replace lots of #hide statements with `@setup` blocks JuliaDocs/Documenter.jl#244
- [x] Fix Theme documentation #852 
- [x] Add Mattriks' new geometry

PR #892, opened 2016-09-07
Fixes #842 + adds alpha for color aesthetic

### LINKED ISSUE #842: Change size of points as a free parameter in plot()?
Is it possible to set the point size to show another dimension of the data in Gadfly? I envision something like the following

``` .julia
plot(x=1:5, y=1:5, color=["a", "b", "b", "a", "c"], size=1:5, Geom.point)
```

Is this possible?

Mattriks:
This PR enables the `size` aesthetic for `Geom.point` (fixes #842), and also enables alpha transparency when using `Geom.point` with the `color` aesthetic. Note that the size aesthetic needs to be of type `Measure`. Example:

``` julia
using RDatasets, DataFrames
msleep = dataset("ggplot2", "msleep")
D = msleep[:,[:Vore,:BrainWt,:BodyWt,:SleepTotal]]
complete_cases!(D)
D[:Sleep] = [ sqrt(x/pi)*mm for x in D[:SleepTotal]  ]

theme(alpha) = Theme(alpha_color=c ->  Colors.RGBA{Float32}(c.r, c.g, c.b, alpha)) 

p = plot(D, x=:BodyWt, y=:BrainWt, size=:Sleep, Geom.point,
    color=:Vore, theme(0.65),
    Scale.x_log10, Scale.y_log10,
    Guide.title("Sleep times of Mammals")
)
```

issue842

### DIFF
--- a/src/geom/point.jl
+++ b/src/geom/point.jl
@@ -42,11 +42,14 @@ function render(geom::PointGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics
 
     aes_x, aes_y = concretize(aes.x, aes.y)
 
+    idcolor = PooledDataArray([color(x) for x in aes.color])
+    alpha_color = PooledDataArray([theme.alpha_color(x) for x in aes.color])
+
     ctx = context()
     if aes.shape != nothing
         xs = Array(eltype(aes_x), 0)
         ys = Array(eltype(aes_y), 0)
-        cs = Array(eltype(aes.color), 0)
+        cs = Array(eltype(alpha_color), 0)
         size = Array(eltype(aes.size), 0)
         shape_max = maximum(aes.shape)
         if shape_max > length(theme.shapes)
@@ -54,7 +57,7 @@ function render(geom::PointGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics
         end
 
         for shape in 1:maximum(aes.shape)
-            for (x, y, c, sz, sh) in Compose.cyclezip(aes.x, aes.y, aes.color,
+            for (x, y, c, sz, sh) in Compose.cyclezip(aes.x, aes.y, alpha_color,
                                                       aes.size, aes.shape)
                 if sh == shape
                     push!(xs, x)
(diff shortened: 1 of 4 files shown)


tlnagy:
Awesome, but the change in transparency values aren't reflected in the legend? Also, I think the alpha function is a bit too complex. It would be better to have something like this:

```julia
Theme(alpha_color = 0.65) 
```
No?

I guess the main other things are to add some tests and docs. I would wait on writing docs until #879 goes through, which should be soon, or you could always rebase on top of that PR.

Mattriks:
Wrt the change in transparency values in the key, I believe that `ggplot2` has a manual `override.aes` argument to change the transparency in the guides (it isn't automated).  I'm not suggesting that is the way to go here!  

The idea of an `alpha_color` function follows the use of other gadfly theme functions e.g. `lowlight_color`. These functions give the user control over both color type and transparency. Implementing `Theme(alpha_color=0.65)` is possible, but then the function it gets passed to might be limited to certain color types. I'll think about it!

I also need to implement a key for the `size` aesthetic.

tlnagy:
> Wrt the change in transparency values in the key, I believe that ggplot2 has a manual override.aes argument to change the transparency in the guides (it isn't automated). I'm not suggesting that is the way to go here! 

I'm of the opinion that the legend should always reflect what is being plotted, but maybe there are cases where that shouldn't be true.

> The idea of an alpha_color function follows the use of other gadfly theme functions e.g. lowlight_color. These functions give the user control over both color type and transparency

I think since the name is already so specific (alpha value change) maybe it would be good to keep it specific (i.e. only change the alpha value)? I would rename is to `alpha` in that case, unless that conflicts with some other `Theme` parameter.

tlnagy:
The PR with all the doc updates has gone through FYI.

tlnagy:
Hey @Mattriks, any chance of fixing this up? I'm aiming for a new Gadfly release soon. I think the new docs, better Juno support, fixes for https://github.com/JuliaStats/DataFrames.jl/pull/1008, and this PR could make a bump to Gadfly version 0.5.0

tlnagy:
Friendly bump. If you get a chance to update this, that would be great.

Mattriks:
I probably should have put these in as separate issues.  To fix #842  only requires the 4 new lines in `scale.jl` and 1 updated line in `statistics.jl` (https://github.com/GiovineItalia/Gadfly.jl/pull/892/files) : can you could update in `master`?  No time atm to work on the alpha issue.

tlnagy:
I packaged your code into a new PR over on #952 so my code from #842 now works as follows:

```
plot(x=1:5, y=1:5, color=["a", "b", "b", "a", "c"], size=map(Measures.Length{:mm, Float64}, 1:5), Geom.point)
```

I might add some convenience functions for Measures to support the `range` operator because the above solution is kinda verbose.

tlnagy:
Actually, I'm not even sure that we should keep using Measures given that https://github.com/ajkeller34/Unitful.jl has taken off.

Mattriks:
For converting to a `Measures` type, these 3 forms will work: 
```Julia
[x*mm for x in 1:5]
map(x->x*mm, 1:5)
(x->x*mm).(1:5)
```
But if you can do it more compactly with a convenience function - great!

tlnagy:
I'm going to go ahead and close this since #952 fixed part of this. I still think having alpha values for the Color aesthetic would be nice so if you package that up in a new PR, I would gladly review it.

PR #893, opened 2016-09-11
Fix use of `invoke`

yuyichao:
* Use tuple type instead of tuples (Ref https://github.com/JuliaLang/julia/pull/18444 and it's easier to infer in general)

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -1018,7 +1018,7 @@ function display(p::Plot)
             @try_display return display(displays[i], p)
         end
     end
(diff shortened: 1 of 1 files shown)


tlnagy:
Any idea why this PR is broken on the nightlies?

yuyichao:
It's unrelated to this PR (also happen on master). Possibly a base bug.

yuyichao:
Test failure is https://github.com/JuliaLang/julia/issues/18449

tlnagy:
Thanks for the sleuthing, @yuyichao. LGTM.

PR #897, opened 2016-09-13
Fix italicized inline code

Hydrotoast:
Double backticks seems to produce italics in the documentation. I think single backticks were intended here.


tlnagy:
Thanks for catching this. Failures on nightly are unrelated.

PR #907, opened 2016-10-04
fixed show methods using Cairo for julia v0.5

tauu:
Try blocks behave differently in julia v0.5 compared to v0.4. The new
behaviour caused the show methods requiring Cairo to be defined in
local scope instead of global scope, which in turn prevented them from
being used. See also:
  https://github.com/JuliaLang/julia/issues/18767

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -969,18 +969,16 @@ end
 
 try
     getfield(Compose, :Cairo) # throws if Cairo isn't being used
-    @compat function show(io::IO, ::MIME"image/png", p::Plot)
+    @compat global show(io::IO, ::MIME"image/png", p::Plot) =
         draw(PNG(io, Compose.default_graphic_width,
-                 Compose.default_graphic_height), p)
-    end
+                 Compose.default_graphic_height), p);
 end
 
 try
     getfield(Compose, :Cairo) # throws if Cairo isn't being used
-    @compat function show(io::IO, ::MIME"application/postscript", p::Plot)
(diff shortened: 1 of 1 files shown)


src/Gadfly.jl
```diff
@@ -969,18 +969,16 @@ end
 
 try
     getfield(Compose, :Cairo) # throws if Cairo isn't being used
-    @compat function show(io::IO, ::MIME"image/png", p::Plot)
+    @compat global show(io::IO, ::MIME"image/png", p::Plot) =
```
tlnagy: Shouldn't this be `global function show`?

tlnagy:
Thanks! I'm waiting for https://github.com/JuliaIO/FileIO.jl/issues/85 to merge before I reboot the tests and make sure everything is okay.

src/Gadfly.jl
```diff
@@ -977,9 +978,10 @@ end
 
 try
     getfield(Compose, :Cairo) # throws if Cairo isn't being used
+    global show
     @compat function show(io::IO, ::MIME"application/postscript", p::Plot)
         draw(PS(io, Compose.default_graphic_width,
-             Compose.default_graphic_height), p)
+             Compose.default_graphic_height), p);
```
tlnagy: Semi-colon is unnecessary, right?

tlnagy:
So  JuliaIO/FileIO.jl#85 is resolved and I rebooted the tests, if they are all green, I'll go ahead and merge this in

PR #908, opened 2016-10-04
RFC: Color schemes

shashi:
This adds the ability to

1. Set color scales in `Theme` (`discrete_color_scale`, `continuous_color_scale`)
2. Ability to set a theme with `set_theme(::Theme)`, or with `set_theme(theme_name::Symbol)` given `get_theme(::Type{Val{theme_name}})` returns a `Theme`
3. A `style` function which will modify the current theme instead of creating a new one with the default default values.
4. Set theme using GADFLY_THEME env variable

and adds a dark theme:

image

image

image

image

image

image

And is intended for use in Atom

image


pkofod:
Is this a theme you cooked up yourself, or is it the atom color scheme, or?

tlnagy:
🎉🎉🎉🎉 

I guess the main things remaining are to update the docs. I think the most important files are `docs/src/man/themes.md` and `docs/src/tutorial.md`. Great work.

src/Gadfly.jl
```diff
@@ -32,6 +32,14 @@ export SVGJS, SVG, PGF, PNG, PS, PDF, draw, inch, mm, cm, px, pt, color, @colora
 function __init__()
     # Define an XML namespace for custom attributes
     Compose.xmlns["gadfly"] = "http://www.gadflyjl.org/ns"
+    if haskey(ENV, "GADFLY_THEME")
+        theme = ENV["GADFLY_THEME"]
+        try
+            set_theme(Symbol(theme))
+        catch err
+            warn("Error loading Gadlfy theme $theme (set by GADFLY_THEME env variable)")
```
tlnagy: Gadfly misspelled

src/Gadfly.jl
```diff
@@ -1132,50 +1140,74 @@ include("statistics.jl")
...
-                                          :ymin       => Scale.y_discrete(),
-                                          :ymax       => Scale.y_discrete(),
-                                          :yintercept => Scale.y_discrete(),
-                                          :xgroup     => Scale.xgroup(),
-                                          :ygroup     => Scale.ygroup(),
-                                          :color      => Scale.color_discrete(),
-                                          :shape      => Scale.shape_discrete(),
-                                          :group      => Scale.group_discrete(),
-                                          :label      => Scale.label()))
+const default_aes_scales = Dict{Symbol, Dict}(
```
tlnagy: Is the `@compat` not necessary?

src/varset.jl
```diff
@@ -47,13 +51,8 @@ macro varset(name::Symbol, table)
...
-            # shallow copy constructor
-            function $(name)(a::$(name))
-                b = new()
-                for name in fieldnames($(name))
-                    setfield!(b, name, getfield(a, name))
-                end
-                b
+            function $(name)($(inherit_parameters_expr), b::$name)
+                $(new_with_defaults)
             end
```
tlnagy: nice!

tlnagy:
Also, it looks like Travis is failing due to things unrelated to this PR. I'm looking at you RData 🙄

src/Gadfly.jl
shashi: Not anymore since we depend on Julia 0.4+

shashi:
@pkofod I kinda made it up...

@tlnagy I've made more changes, now it we have a `theme_stack` where you can push and pop themes from as you call `plot` to apply different themes. See the documentation... I have some questions about it, I will be making some inline comments.

docs/src/man/themes.md
```diff
@@ -82,13 +95,51 @@ srand(12345)
...
+
+Same effect can be had with `with_theme`
+
+```@example 1
+Gadfly.with_theme(dark_panel) do
+
+  plot(x=rand(10), y=rand(10),
+       style(major_label_font="CMU Serif",minor_label_font="CMU Serif",
+             major_label_font_size=16pt,minor_label_font_size=14pt))
+end
```
shashi: How do I hide the output of this? (This is the same as the previous one)

docs/src/man/themes.md
```diff
@@ -82,13 +95,51 @@ srand(12345)
...
+
+```@example 1
+
+Gadfly.push_theme(dark_panel)
+
+Gadfly.with_theme(
+       style(major_label_font="CMU Serif",minor_label_font="CMU Serif",
+             major_label_font_size=16pt,minor_label_font_size=14pt)) do
+
+  plot(x=rand(10), y=rand(10))
```
shashi: Need to hide this as well.

ViralBShah:
That color scheme will look really awesome with Juno!

src/Gadfly.jl
tlnagy: Ah, good point. I forgot we dropped 0.3

src/statistics.jl
```diff
@@ -1492,7 +1492,7 @@ const contour = ContourStatistic
 function default_scales(::ContourStatistic)
     return [Gadfly.Scale.z_func(), Gadfly.Scale.x_continuous(),
             Gadfly.Scale.y_continuous(),
-            Gadfly.Scale.color_continuous_gradient()]
+            Gadfly.get_scale(:numerical, :color)]
```
tlnagy: Yeah, shouldn't these all be `get_scale` calls?

docs/src/man/themes.md
tlnagy: As in, you don't want it to plot? You can append ` # hide` (spaces are important) to the end of any line you don't want displayed. If you have multiple lines then a `@setup` block might be useful. See https://juliadocs.github.io/Documenter.jl/latest/man/syntax.html#@example-block-1

tlnagy:
I think this PR is getting close. Main thing is to get https://github.com/JuliaIO/FileIO.jl/issues/85 fixed because I want CI tests passing before merging.

src/statistics.jl
shashi: I'll leave the others unchanged for now, because its seems this will in fact be faster than a dict lookup. We should consider this if we are doing a larger simplification of these things.

docs/src/man/themes.md
shashi: I want the code to appear, but not the output...

tlnagy:
The tests look good! (except for on 0.4, but that's due to Media/Juno having issues)

docs/src/man/themes.md
shashi: Ah, thanks. I think 1st one will also test that it works, so will use that.

shashi:
Okay, added a couple of tests. Added a bit more documentation. I did a Documenter generation and everything is great. Merging.

PR #920, opened 2016-10-15
Juno fixes

MikeInnes:
I've altered the default theme setting to work to be in line with what PlotlyJS.jl does. We check if we're booted inside Juno and set the theme accordingly there.

I also removed the white background from the Gadfly render output, which is no longer necessary.

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -1134,13 +1134,13 @@ import Juno: Juno, @render, media, Media, Hiccup
 media(Plot, Media.Plot)
 
 @render Juno.PlotPane p::Plot begin
-  x, y = Juno.plotsize()
-  set_default_plot_size(x*Gadfly.px, y*Gadfly.px)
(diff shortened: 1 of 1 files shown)


src/Gadfly.jl
```diff
@@ -1134,8 +1136,7 @@ media(Plot, Media.Plot)
 @render Juno.PlotPane p::Plot begin
   x, y = Juno.plotsize()
   set_default_plot_size(x*Gadfly.px, y*Gadfly.px)
-  Hiccup.div(Dict(:style=>"background: white"),
-                  HTML(stringmime("text/html", p)))
+  HTML(stringmime("text/html", p))
```
tlnagy: I must've missed this last time. Could you use 4 spaces for indentation here?

MikeInnes: Good catch, fixed that up

tlnagy:
Thanks!

PR #923, opened 2016-10-17
Hide API pages in the sidebar

mortenpi:
Use the new `hide` feature of Documenter to hide a bunch of pages in the sidebar. Also, move the index pages to the Library section since they feel more like an API reference rather than a manual page.

Demo: http://mortenpi.eu/Gadfly.jl/docs-hidden-pages/

If you'd like to keep the Geometries etc. pages under Manual, I'd be happy to make the change, just this way felt better to me.

**Important:** Documenter `v0.6.0` has to be published on METADATA (JuliaLang/METADATA.jl#6776) before this can be merged.


tlnagy:
Looks great. Thanks for doing this!

PR #933, opened 2016-11-28
document display() and trailing semi-colons on plot

bjarthur:
i recently became confused as to why plot were not being interactively displayed when generated within functions.  thought new users would benefit from more documentation.


tlnagy:
Thanks! I agree that this could be confusing for users using gadfly interactively.

PR #934, opened 2016-11-28
add methods to display h/v stacks

bjarthur:
i'm not super familiar with the Gadfly codebase, or even intended API, but i find displaying plots in the browser super useful.  this PR adds the ability to display stacks as well.

### DIFF
--- a/docs/src/man/layers.md
+++ b/docs/src/man/layers.md
@@ -57,8 +57,8 @@ plot details than is available with [Geom.subplot_grid](@ref).
 ```julia
 p1 = plot(x=[1,2,3], y=[4,5,6])
 p2 = plot(x=[1,2,3], y=[6,7,8])
-draw(PDF("p1and2.pdf", 6inch, 6inch), vstack(p1,p2))
+vstack(p1,p2)
 p3 = plot(x=[5,7,8], y=[8,9,10])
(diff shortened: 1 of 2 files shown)


docs/src/man/layers.md
```diff
@@ -57,8 +57,8 @@ plot details than is available with [Geom.subplot_grid](@ref).
 ```julia
 p1 = plot(x=[1,2,3], y=[4,5,6])
 p2 = plot(x=[1,2,3], y=[6,7,8])
-draw(PDF("p1and2.pdf", 6inch, 6inch), vstack(p1,p2))
+vstack(p1,p2)
 p3 = plot(x=[5,7,8], y=[8,9,10])
 p4 = plot(x=[5,7,8], y=[10,11,12])
-draw(PDF("p1to4.pdf", 6inch, 9inch), vstack(hstack(p1,p2),hstack(p3,p4)))
+vstack(hstack(p1,p2),hstack(p3,p4))
```
tlnagy: Don't replace the PDF generation. I would probably make a separate test for displaying stacks.

tlnagy: Yeah, you're right. Not sure why that test was that way.

tlnagy:
I'll merge once the tests pass.

Issue #600, opened 2015-06-01
Titles after vstack/hstack


tcovert:
Is it possible to add titles to a grid of figures?  Say I've got plots p1-p4, and I combine them and render to a PDF (as in the documentation):
draw(PDF("file.pdf",6inch,6inch),vstack(hstack(p1,p2),hstack(p3,p4))

I see that its possible to add titles to each of the individual plots, but how would I add a title above (or below) the grid of p1-p4?

dcjones:
There's not any special support for this, but it's possible to do "manually" through Compose:
```julia
p = hstack(plot(x=rand(50), y=rand(50), Geom.point),
           plot(x=rand(50), y=rand(50), Geom.point))
title = compose(context(0, 0, 1w, 0.25inch),
                text(0.5, 1.0, "These are some plots", hcenter, vbottom))
vstack(title, p)
```

![try](https://cloud.githubusercontent.com/assets/210361/7916207/434ee26a-0837-11e5-8d3e-9b1074d6bbdd.png)

tcovert:
Thanks, this is great.  I assume I need to add "using Compose" to the top of my code?

dcjones:
Oh, right. You do need `using Compose`.

tcovert:
is there documentation for the compose() and text() functions somewhere?  I didn't see it on composejl.org

IainNZ:
Those are the docs right now, plus the source itself
> The all-important function in Compose, is called, not surprisingly, compose. Calling `compose(a, b)` will return a new tree rooted at a and with b attached as a child.

dcjones:
There's unfortunately not API documentation for things like `text`.

protogeezer:
What's the best way to handle the case of an odd number of plots? In other words, how would one leave a blank on one side of the "bottom" row?

protogeezer:
FWIW, here is my best shot. I couldn't figure out how to specify font attributes other than family and size. Creating an all white plot is a bit crude, but it does seem to work.
```
        title1 = compose(context(0, 0, 1w, 0.2h),font("Times"),fontsize(24pt), text(0.5, 1.0, 
                                     "The first line of the title", hcenter, vbottom))
	title2 = compose(context(0, 0, 1w, 0.05h),font("Helvetica"),fontsize(14pt), text(0.5, 1.0, 
                                     "This is the stuff you don't want anyone to notice", hcenter, vbottom))
[...]

PR #937, opened 2016-11-30
title(), gridstack()

bjarthur:
closes https://github.com/GiovineItalia/Gadfly.jl/issues/600

specifically, this PR adds `title` which puts a string at the top of a `vstack` or `hstack`, as described in the issue above.

a second commit, which i could put into a separate PR if you like, documents the already existing yet not exported `gridstack`.  i took the liberty of exporting it.

lastly, i changed the "Stacks and Layers" header to "Layers and Stacks" to match the order of the subsequent subsections.

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -24,7 +24,7 @@ export Plot, Layer, Theme, Col, Scale, Coord, Geom, Guide, Stat, render, plot,
 
 
 # Re-export some essentials from Compose
-export SVGJS, SVG, PGF, PNG, PS, PDF, draw, inch, mm, cm, px, pt, color, @colorant_str, vstack, hstack
+export SVGJS, SVG, PGF, PNG, PS, PDF, draw, inch, mm, cm, px, pt, color, @colorant_str, vstack, hstack, title
 
 
 function __init__()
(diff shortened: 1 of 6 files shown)


src/Gadfly.jl
```diff
@@ -934,6 +934,14 @@ function draw(backend::Compose.Backend, p::Plot)
...
 end
 
+"""
+    title(str::String, ctx::Context) -> Context
+
+Add a title string to a group of plots, typically created with `vstack`, `hstack`, or `gridstack`.
+"""
+title(str::String, ctx::Context) = vstack(
+    compose(context(0, 0, 1, 0.1), text(0.5, 1.0, str, hcenter, vbottom)),
+    compose(context(0, 0, 1, 0.9), ctx))
```
tlnagy: Missing the closing `end`. This is causing tests to fail.

tlnagy:
I appreciate the documentation update and the exporting of `gridstack`. I'm a bit torn about `title()` because it's not very grammar-of-graphics-y. Maybe allowing `vstack` and `hstack` to take `Geom.title` would be better? @shashi

bjarthur:
i see your point re. considering inputing `Geom.title` to `vstack` and `hstack` instead.  i'm also a bit torn though because the stack interface in itself is already not very grammer-of-graphics-y.  were it not for `Geom.subplot_grid` i'd suggest that said grammer be restricted to just the creation of a single pair of axes.

tlnagy:
Alright. After thinking about this, I think we should merge this. I agree with your assessment that the whole stack approach isn't very GoG-esque.

test/REQUIRE
```diff
@@ -1,2 +1,3 @@
 RDatasets
-Cairo
\ No newline at end of file
+Cairo
+Compose
```
tkelman: Compose is already in the main `REQUIRE` so not necessary here

PR #943, opened 2017-01-02
Update docs to show use of the shape aesthetic

zacsketches:
The latest version of the docs had a todo here.  Since I was using the shape functionality in a plot I'm working on tonight, I thought I'd go ahead and update the docs too.  Hope this helps.  Use of the shape aesthetic works on my Julia v0.5.0.


tlnagy:
Yeah, the breakage on nightly is unrelated. Thanks for doing this!

Issue #278, opened 2014-04-26
abline geom


diegozea:
In addition to hline and vline, an abline geom [(like the abline of ggplot2)](http://docs.ggplot2.org/0.9.3.1/geom_abline.html) would be great! ;)

bjarthur:
fixed by https://github.com/GiovineItalia/Gadfly.jl/pull/957.  this issue can be closed i believe.

Issue #79, opened 2013-09-26
Horizontal/Vertical lines


Keno:
It would be nice if one could add vertical and horizontal (dashed ideally) guides/lines to a plot at arbitrary positions. If this is possible already and I somewhat missed it, consider this a doc issue ;).

PR #957, opened 2017-02-20
add abline Geometry

bjarthur:
handy to be able to plot `y=x` easily.


bjarthur:
test pass on julia 0.5 but not nightly for this PR.

but...  why on the gadfly.jl home page does it say the package evaluator julia 0.5 tests are failing for gadfly 0.5.3?

also, can someone give me a primer on why `src/precompile.jl` is necessary?  shouldn't the call to `__precompile__()`  in `src/Gadfly.jl` suffice?  i sprinkled a few uses of `Geom.yequalsx` in there, but coverage might not be complete.

docs/src/lib/geoms/geom_yequalsx.md
```diff
@@ -0,0 +1,25 @@
+```@meta
+Author = "Ben J. Arthur"
+```
+
+# Geom.hline
```
tlnagy: Shouldn't this be Geom.yequalsx?

tlnagy:
I'm not sure why we have `precompile.jl`, but Gadfly is pretty old for a Julia package so it might something left over before packages were precompiled automatically starting in 0.4

Mattriks:
What about developing this into `Geom.abline` (#278) of which `Geom.yequalsx` would be a particular case.

tlnagy:
Ah, I thought we already had that Geom, but since we don't, I think that would be best. As it is currently, `Geom.yequalsx` seems to specific. Something like `Geom.abline` would be more general and more useful.

bjarthur:
great idea.  would it then make sense to get rid of `hline` and `vline` since they are special cases of `abline`?

Mattriks:
The coding of `Geom.hline` and `Geom.vline` is a bit different: `Geom.hline` and `Geom.vline` both use absolute and relative plot coordinates, whereas `Geom.yequalsx` uses absolute (not relative) coordinates. `Geom.abline` I think would also use absolute coordinates.

tlnagy:
It might be nice to have `Geom.hline` and `Geom.vline` become aliases for special cases of `Geom.abline`. I'm not sure having the relative plot coordinates is necessary since they aren't mentioned in the docs and AFAIK [untested](https://github.com/GiovineItalia/Gadfly.jl/blob/master/test/hline_vline.jl) as well

Mattriks:
Relative coordinates are in the docs for `Compose` (and not `Gadfly`), because it is a very useful feature for graphic developers. Gadfly users don't need to know about relative coordinates.

tlnagy:
Yeah, I see that here: https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/geom/hline.jl#L33

I would imagine that `Geom.abline` might have to use relative units as well to make sure it extends all the way through the whole visible plot area. No?

tlnagy:
> but... why on the gadfly.jl home page does it say the package evaluator julia 0.5 tests are failing for gadfly 0.5.3?

Also, @bjarthur, I think this is due to some problems with Fontconfig on the vagrant build used to do the nightly tests.

bjarthur:
but it's reported that it's failing for 0.5, not the nightly

tlnagy:
I meant the nightly builds of 0.5. PackageEvaluator.jl attempts to build each package on each version of Julia every night. It looks like something may have broke with it. I'm discussing the problem with Tony: https://github.com/JuliaCI/PackageEvaluator.jl/issues/153

bjarthur:
vertical lines can't be drawn with `y=mx+b`.  but there still is a nice way to consolidate `hline` and `vline` into `abline`.  what would you think adding `xslope` and `yslope` aesthetics to the existing `xintercept` and `yintercept`.  then the `abline` geometry draws a `y = xslope * x + yintercept` if those aesthetics exist, and an `x = yslope * y + xintercept` if those aesthetics exists, or both if they both do.  so horizonal, vertical, and sloped lines can all be drawn at once with a single consolidated geometry.  make sense?

bjarthur:
tests are passing on what i had in mind re. deprecating `hline` and `vline` and folding that functionality into the new `abline`.  please let me know what you think.

also, is there any reason not to have travis test on OS X?  i think that'd be a simple change to `travis.yml`.  not sure about testing on windows.

tlnagy:
This looks great! I'm not sure that we should get rid of `hline` and `vline`, would it be possible to have them be aliases for special cases of abline? Something like what we do for `Geom.path`? See https://github.com/GiovineItalia/Gadfly.jl/blob/1bcd9fb335d9f25d1edd8c810d9f576b582caa26/src/geom/line.jl#L31-L33. 

> also, is there any reason not to have travis test on OS X? i think that'd be a simple change to `travis.yml`. not sure about testing on windows.

I disabled testing on OS X in e6ea3310cf4cf95ebe507c9b4991e54097ddad17 because Travis' OS X build bots are underprovisioned and they would often stall out or take forever to run. I would be open to running tests on Windows though via appveyor.

Mattriks:
1. Yes I also think that `Geom.hline` and `Geom.vline` should either be retained, or written as particular cases of `Geom.abline`. 
2. Color. There seems to be three ways to color a geometry: 
- `Geom._(color=)` e.g. used by `hline` and `vline`
- `layer(Geom._ , Theme(default_color=))`, works generally
- `layer(Geom._, color=Vector{Color})` e.g. [`Geom.segment`](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/geom/segment.jl) allows this, which means different lines can have different colors
My question is should `Geom.abline` also allow the third way?  The first and second way seem to do the same thing. There was probably a reason that `hline` and `vline` were coded with the first way #79  (anybody know)?

bjarthur:
i've re-instated `hline` and `vline` as special cases `abline`.

re. color, i agree a vector is useful, and have added it as an argument.  doing so as an aesthetic would not permit `abline` to be combined with any other geometry which also used the color aesthetic.  i guess this is why it was originally done this way for `h/vline`.

src/aesthetics.jl
```diff
@@ -169,7 +171,7 @@ function assert_aesthetics_undefined(who::AbstractString, aes::Aesthetics, vars:
     defined_vars = intersect(Set(vars), defined_aesthetics(aes))
     if !isempty(defined_vars)
         error(@sprintf("The following aesthetics are defined but incompatible with %s: %s\n",
-                       who, join(undefined_vars, ", ")))
+                       who, join(defined_vars, ", ")))
```
tlnagy: Why make this change?

bjarthur: because `undefined_vars` is not defined in that function.  calling it throws an error.  it's clearly a typo.  compare to the function immediately preceding this one.

also, let's hold off on merging for a couple days just to make sure all is well.  i still, for example, want to understand why `src/precompile` is necessary.  Gadfly and Compose are the only packages i've seen that do this.

tlnagy: Ah. Nice catch!

I looked into the `precompile.jl` code and I was convinced that it's a relic of pre-0.4 julia. I ripped it all out and as far as I can tell it doesn't introduce any regressions and introduces a dramatic speed in precompilation. I have a PR here: #958

tlnagy:
Also, one last thing, can you add this to `NEWS.md`? I'm trying to get us back to using it for new releases.

tlnagy:
Can you rebase this? LGTM.

bjarthur:
rebased and passing on 0.5

tlnagy:
LGTM 🎉

Issue #961, opened 2017-02-24
Geom.hline issue with Dates


Mattriks:
This occurs on master. The plot works if you remove hline from the plot statement.
```julia
day = collect(Date("1960-01-01"):Dates.Day(1):Date("1999-12-31"))
t = Dates.value.(day)
w = 2π/365.25
D1 = DataFrame(Day=day, y=2*rand(length(day)).*(1+sin(w*t)))
hline = Geom.hline(color="red")
p = plot(D1, x=:Day, y=:y, yintercept=[3.9], Geom.line, hline )
```
resulting in the error:
```julia
MethodError: no method matching isless(::Float64, ::Date)
Closest candidates are:
  isless(::Float64, ::Float64) at float.jl:283
  isless(::Date, ::Date) at dates\types.jl:239
  isless(::DataArrays.NAtype, ::Any) at C:\Users\Matt\.julia\v0.5\DataArrays\src\operators.jl:510
  ...

 in min(::Date, ::Float64) at .\operators.jl:88
 in render(::Gadfly.Geom.ABLineGeometry, ::Gadfly.Theme, ::Gadfly.Aesthetics) at C:\Users\Matt\.julia\v0.5\Gadfly\src\geom\abline.jl:78
```

bjarthur:
oi, my bad.  see new PR.

PR #962, opened 2017-02-24
revert to original {h,v}line, adding color/size vectors

bjarthur:
closes https://github.com/GiovineItalia/Gadfly.jl/issues/961

handling aesthetics of mixed types when trying to unify h,v,abline code, so i split them back out, but kept them in the same file at least, and retained the feature to make multiple lines different colors/sizes.


tlnagy:
Nice!

Issue #770, opened 2015-12-15
boxplot fails with ERROR: Vector form and vector property differ in length


bastinat0r:
I am using julia to generate plots for my thesis. 
Currently I am working on a quite huge [dataset](http://bastinat0r.de/nightshift.tar.gz) of that I can only plot smaller subsets without errors.
When trying to draw the plot for the whole data I get the following error:

```
Error showing value of type Gadfly.Plot:
ERROR: Vector form and vector property differ in length. Can't distribute.
 in print_vector_properties at /home/sebastian/.julia/v0.4/Compose/src/svg.jl:759
 in draw at /home/sebastian/.julia/v0.4/Compose/src/svg.jl:1200
 in drawpart at /home/sebastian/.julia/v0.4/Compose/src/container.jl:548
 in drawpart at /home/sebastian/.julia/v0.4/Compose/src/container.jl:586
 in drawpart at /home/sebastian/.julia/v0.4/Compose/src/container.jl:580
 in drawpart at /home/sebastian/.julia/v0.4/Compose/src/container.jl:586 (repeats 3 times)
 in drawpart at /home/sebastian/.julia/v0.4/Compose/src/container.jl:580
 in drawpart at /home/sebastian/.julia/v0.4/Compose/src/container.jl:586
 in drawpart at /home/sebastian/.julia/v0.4/Compose/src/container.jl:473
 in drawpart at /home/sebastian/.julia/v0.4/Compose/src/container.jl:586 (repeats 2 times)
 in draw at /home/sebastian/.julia/v0.4/Compose/src/container.jl:430
 in display at /home/sebastian/.julia/v0.4/Gadfly/src/Gadfly.jl:963
 [inlined code] from multimedia.jl:151
 in display at /home/sebastian/.julia/v0.4/Gadfly/src/Gadfly.jl:922
 in print_response at REPL.jl:134
 in print_response at REPL.jl:121
 in anonymous at REPL.jl:624
```

What I want to do is basically draw a boxplot showing the statistical relation between range (measurement) and distance (real value).

```
function bucketplot(df, bsize)
        df[:bucket] = bsize * div( df[:distance], bsize ) + bsize / 2
        return plot(df, x=:bucket, y=:range, Geom.boxplot)
end
```

However plotting fails with the error I mentioned. Is it possible that the plots are not scaled to the same dimension? Should I use different asthetics to generated the plot?
I tried to find some other description of the problem, but I seem to be quite alone with this.

[...]

PR #963, opened 2017-02-25
temporary fix for issue 770

bastinat0r:
As discussed in https://github.com/GiovineItalia/Gadfly.jl/issues/770 if there are more then 100 outlies the boxplot fails with a hard to understand error message.

The changes made mitigate this problem by not drawing the outliers and shows a warning for the user to explain why this happens.


tlnagy:
See comment in #770

PR #964, opened 2017-03-03
fixup density2d

bjarthur:
small typo computing the bandwidth of the kernels used causes errors when the ranges of x and y are wildly different.


tlnagy:
Nice catch

PR #973, opened 2017-04-07
WIP:  update for julia 0.6

bjarthur:
tests are still failing to a macro call in Compose:

```
while loading /Users/arthurb/.julia/v0.6/Gadfly/test/runtests.jl, in expression starting on line 202
FAILED!
ERROR: LoadError: UndefVarError: codes not defined
Stacktrace:
 [1] macro expansion at /Users/arthurb/.julia/v0.6/Compose/src/misc.jl:125 [inlined]
 [2] jscall(::Array{String,1}, ::Array{Array{Measures.Measure,1},1}) at /Users/arthurb/.julia/v0.6/Compose/src/property.jl:587 (repeats 2 times)
 [3] (::Gadfly.Guide.##3#11{Array{Int64,1},Array{ColorTypes.Color,1},DataStructures.OrderedDict{ColorTypes.Color,AbstractString},Gadfly.Scale.#labeler#66,Compose.Context,Gadfly.Theme,Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}})(::Compose.ParentDrawContext) at /Users/arthurb/.julia/v0.6/Gadfly/src/guide.jl:277
```

can anyone give me a hint here?

there is still also the `show` / `display` code to port.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.5
 Colors 0.3.4
 Compat 0.8.5
-Compose 0.4.4
+Compose 0.5
 Contour 0.1.1
 DataFrames 0.4.2
 DataStructures
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -44,7 +44,7 @@ function __init__()
 end
 
 
-typealias ColorOrNothing @compat(Union{Colorant, (@compat Void)})
+const ColorOrNothing = @compat(Union{Colorant, (@compat Void)})
 
 element_aesthetics(::Any) = []
 input_aesthetics(::Any) = []
@@ -55,12 +55,12 @@ default_statistic(::Any) = Stat.identity()
 element_coordinate_type(::Any) = Coord.cartesian
 
 
-abstract Element
-abstract ScaleElement       <: Element
-abstract CoordinateElement  <: Element
-abstract GeometryElement    <: Element
-abstract GuideElement       <: Element
-abstract StatisticElement   <: Element
+@compat abstract type Element end
+@compat abstract type ScaleElement       <: Element end
+@compat abstract type CoordinateElement  <: Element end
+@compat abstract type GeometryElement    <: Element end
+@compat abstract type GuideElement       <: Element end
+@compat abstract type StatisticElement   <: Element end
 
 
 include("misc.jl")
@@ -77,7 +77,7 @@ include("theme.jl")
 
 # The layer and plot functions can also take functions that are evaluated with
 # no arguments and are expected to produce an element.
-typealias ElementOrFunction{T <: Element} @compat(Union{Element, Base.Callable, Theme})
+@compat const ElementOrFunction{T <: Element} = @compat(Union{Element, Base.Callable, Theme})
 
 const gadflyjs = joinpath(dirname(Base.source_path()), "gadfly.js")
 
@@ -184,7 +184,7 @@ end
 
 
 function add_plot_element!(lyrs::Vector{Layer}, arg::GeometryElement)
-    if ! is(lyrs[end].geom, Geom.nil())
+    if lyrs[end].geom !== Geom.nil()
         push!(lyrs, copy(lyrs[end]))
     end
     lyrs[end].geom = arg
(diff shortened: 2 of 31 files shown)


tlnagy:
That's not a very helpful error message...

tlnagy:
Somehow I think that `codes` is not passing properly to this function and then the inner macro call. No idea why that is happening now. https://github.com/GiovineItalia/Compose.jl/blob/7ea1616cc084e6426bb5a6db422d564328cd6edd/src/property.jl#L585-L591

bjarthur:
all tests pass now.

there are still some deprecation warnings though, mostly from FixedSizeArrays.jl, which Contour.jl uses.    we should consider switching to StaticArrays.jl at some point.  difficulty there is that it has dropped support for 0.5.

Media.jl also throws a couple warnings.

i would still suggest not merging this PR until folks have actually used it.  i have not actually visually checked the appearance of any plots yet to make sure they look okay.

bjarthur:
wait, nevermind, the `codes not defined` error was reared it's head again...

bjarthur:
the `codes not defined` error was a macro hygiene problem.  not sure how it ever worked on julia 0.5.  adding a few `esc()` got it working on 0.6.

now the tests are stuck on a world age problem...

tlnagy:
Awesome! Thanks for finding that out. There seem to be a couple bugs in 0.5 that Gadfly/Compose relied on that were fixed in 0.6

bjarthur:
the world age problem is now fixed, and all tests pass locally on julia 0.5 and 0.6!

would be great if someone could give it a try.  you'll need to checkout master on Loess, Hexagons, and Showoff, and checkout bja/julia06 on Compose and Gadfly.

there are still a couple depwarns from the Media.jl dependency, which is for Juno.

and a dozen or so depwarns remain from FixedSizeArrays.jl, which Contour.jl uses.  the plan is not to ever upgrade FixedSizeArrays.jl to julia 0.6, but rather to migrate everything that depends on it to StaticArrays.jl.

there are also quite a few `broadcast` depwarns, and a handful from `cycle`.  the file/line# info is not specific enough for me fix them.  i'm guessing they come from some dependency, perhaps FixedSizeArrays.

at any rate, as soon as my Compose PR is merged and tagged, the Gadfly tests should pass for this PR, and i think we can merge.

tlnagy:
Is StaticArrays a drop in replacement for FixedSizeArrays? Because then we could move over to StaticArrays for both 0.5 and 0.6. Otherwise, we might have play around with conditional dependencies or something.

bjarthur:
the names have changed, but otherwise it's drop-in.  whenever someone gets around to migrating Contour from FixedSizeArrays to StaticArrays, Gadfly will either have to drop support for julia 0.5 or add an upper bound on the Contour version.  until then, we just have to live with a few deprecation warnings.

tlnagy:
StaticArrays appears to have 0.5 as the lower bound so we could probably migrate Contour right now.

bjarthur:
[this](https://github.com/JuliaArrays/StaticArrays.jl/blob/master/REQUIRE) says 0.6-.  also, Contour.jl is in the [process](https://github.com/tlycken/Contour.jl/issues/37) of being moved to JuliaGeometry.

tlnagy:
Ah, I missed that. I just read the README, but I guess that was out of date. I guess the moment that 0.6 drops we can make a final release for 0.5 and then move master solely to 0.6 and switch to staticarrays.

ararslan:
Does anything else need to happen before you guys can make a 0.6-compatible tag of Gadfly?

bjarthur:
someone should check that it works in the various IDEs... Juno, IJulia, JuliaBox.  i've only tested on the REPL.  specifically, Media.jl throws a few dep warns.

ararslan:
IIRC the version of 0.6 on JuliaBox is pretty old so whether Gadfly works there is probably not entirely representative. If I may, I'd recommend that if it works in the REPL you'd might as well tag a version since mostly works is better than not working at all, and as Tony says, "tags are cheap."

bjarthur:
@tlnagy @Mattriks either of you have any plans to submit PRs in the imminent future?  if not, we should tag a release.

Mattriks:
Yes I hope to submit `Geom.vectorfield` (#608 & #791) say next week sometime.

tlnagy:
As @tkelman says, tagging is cheap. Lets go ahead and go forward with tagging a release so we have 0.6 compat.

Issue #822, opened 2016-04-08
Histogram density y-values wrong in some cases


tlnagy:
For example, the following breaks the computation of the y-values:

``` .julia
using Gadfly
using Distributions
vals = rand(Normal(0, 1), 10000)
p = plot(x=vals/sum(vals), Geom.histogram(density=true))
draw(PNG("densityproblem.png", 20cm, 12cm), p)
```

![densityproblem](https://cloud.githubusercontent.com/assets/1661487/14393422/d57f5bc6-fd7b-11e5-8507-5f920e93e916.png)

tlnagy:
This seems to be related to #574 and #781

scidom:
Has there been any update on this?

tlnagy:
None, sorry. I would be happy to review any PR if you want to tackle this, but I haven't had the time to track this one down.

Jacob640:
I think this problem is due to the pull request that was introduced for #880. The fix replaced  the simultaneous assignment and element-wise division with separate operations thereby changing the precedence of the expression 

`bincounts ./= sum(bincounts) * binwidth`
to
`bincounts = bincounts ./ sum(bincounts) * binwidth`

Adding the brackets seems to do the trick

`bincounts = bincounts ./ (sum(bincounts) * binwidth)`

`plot(x=rand(Normal(), 1000), Geom.density, Geom.histogram(density=true))`
![image](https://cloud.githubusercontent.com/assets/1777669/22694310/9bf525b2-ed3e-11e6-9053-6cd671a78671.png)


I'll submit a pull request later

tlnagy:
Thanks for looking into this! I'll be happy to review and merge if you submit a PR.

bjarthur:
closed by https://github.com/GiovineItalia/Gadfly.jl/pull/977

PR #977, opened 2017-04-12
Fix Bug #822

trohwer:
I did not directly test on trunk, but made the same change on my local version, checking that it fixes the problem. Since it is a relative simple change without side effects, this is probably acceptable.

### DIFF
--- a/src/statistics.jl
+++ b/src/statistics.jl
@@ -338,7 +338,7 @@ function apply_statistic(stat::HistogramStatistic,
             x_min = Gadfly.concrete_minimum(values)
             span = x_max - x_min
             binwidth = span / d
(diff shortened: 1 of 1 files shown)


tlnagy:
Thanks! We'll wait for the tests and then go ahead and merge.

tlnagy:
Hey @bjarthur, any idea why the DateTime test is failing now? It's failing for me locally too, but it passed on master a week ago.

tlnagy:
It might be due to some changes to Measures.jl

bjarthur:
this is fixed in my 0.6 Gadfly PR.  just wait for it to merge.

bjarthur:
re-doing travis

bjarthur:
thanks for this @trohwer !

PR #978, opened 2017-04-12
coerce loess input to vector of floats

### LINKED ISSUE #975: Error with Geom.smooth when using datatype Datetime for x values
Hello,

the above situation yields the following trace:

ERROR: MethodError: no method matching predict(::Loess.LoessModel{Float64}, ::Array{DateTime,1})
Closest candidates are:
  predict{T<:AbstractFloat}(::Loess.LoessModel{T<:AbstractFloat}, !Matched::T<:AbstractFloat) at /home/tr/.julia/v0.5/Loess/src/Loess.jl:129
  predict{T<:AbstractFloat}(::Loess.LoessModel{T<:AbstractFloat}, !Matched::AbstractArray{T<:AbstractFloat,1}) at /home/tr/.julia/v0.5/Loess/src/Loess.jl:134
  predict{T<:AbstractFloat}(::Loess.LoessModel{T<:AbstractFloat}, !Matched::AbstractArray{T<:AbstractFloat,2}) at /home/tr/.julia/v0.5/Loess/src/Loess.jl:167
 in apply_statistic(::Gadfly.Stat.SmoothStatistic, ::Dict{Symbol,Gadfly.ScaleElement}, ::Gadfly.Coord.Cartesian, ::Gadfly.Aesthetics) at /home/tr/.julia/v0.5/Gadfly/src/statistics.jl:1222
 in apply_statistics(::Array{Gadfly.StatisticElement,1}, ::Dict{Symbol,Gadfly.ScaleElement}, ::Gadfly.Coord.Cartesian, ::Gadfly.Aesthetics) at /home/tr/.julia/v0.5/Gadfly/src/statistics.jl:38
 in render_prepare(::Gadfly.Plot) at /home/tr/.julia/v0.5/Gadfly/src/Gadfly.jl:760
 in render(::Gadfly.Plot) at /home/tr/.julia/v0.5/Gadfly/src/Gadfly.jl:819
 in draw(::Compose.Image{Compose.CairoBackend}, ::Gadfly.Plot) at /home/tr/.julia/v0.5/Gadfly/src/Gadfly.jl:934

The reason is, that in statistics.jl around line 1222 the vector aes.x is passed unconverted to Loess.predict.
Changing the code (for the loess case) to

        aes.x = collect((x_min + nudge):((x_max - x_min) / num_steps):(x_max - nudge))

        if stat.method == :loess
            aes.y = Loess.predict(loess(xs, ys, span=stat.smoothing), convert(Vector{Float64}, aes.x))

,i.e. converting aes.x before passing to predict, fixes the problem for me.
There are some other places, where similar adaptations need to be done for other cases.

tlnagy:
fixes #975

### DIFF
--- a/src/statistics.jl
+++ b/src/statistics.jl
@@ -1219,7 +1219,7 @@ function apply_statistic(stat::SmoothStatistic,
         aes.x = collect((x_min + nudge):((x_max - x_min) / num_steps):(x_max - nudge))
 
         if stat.method == :loess
-            aes.y = Loess.predict(loess(xs, ys, span=stat.smoothing), aes.x)
+            aes.y = Loess.predict(loess(xs, ys, span=stat.smoothing), Float64.(aes.x))
         elseif stat.method == :lm
             lmcoeff = linreg(xs,ys)
             aes.y = lmcoeff[2].*aes.x .+ lmcoeff[1]


src/statistics.jl
```diff
@@ -1219,7 +1219,7 @@ function apply_statistic(stat::SmoothStatistic,
         aes.x = collect((x_min + nudge):((x_max - x_min) / num_steps):(x_max - nudge))
 
         if stat.method == :loess
-            aes.y = Loess.predict(loess(xs, ys, span=stat.smoothing), aes.x)
+            aes.y = Loess.predict(loess(xs, ys, span=stat.smoothing), convert(Vector{Float64}, aes.x))
```
bjarthur: julia 0.6 wants `Float64.(aes.x)` instead of `convert...`

tlnagy: Good point. We'll probably want to throw a `@compat` on there as well

bjarthur: i don't think `@compat` is needed for that particular use of the `.` operator.

tlnagy: Okay. I got rid of it.

src/statistics.jl
```diff
@@ -1219,7 +1219,7 @@ function apply_statistic(stat::SmoothStatistic,
         aes.x = collect((x_min + nudge):((x_max - x_min) / num_steps):(x_max - nudge))
 
         if stat.method == :loess
-            aes.y = Loess.predict(loess(xs, ys, span=stat.smoothing), aes.x)
+            aes.y = Loess.predict(loess(xs, ys, span=stat.smoothing), Float64.(aes.x))
```
ararslan: When `aes.x` is a `DateTime` as in the linked issue, this will produce deprecation warnings on Julia 0.6. The proper way to convert a `DateTime` to a float is to call `Dates.value` and then convert.

One way to approach this would be to define some helper function that can handle dates appropriately:
```julia
@inline _aesconvert{T<:Real}(::Type{T}, d::DateTime) = convert(T, Dates.value(d))
@inline _aesconvert{T<:Real}(::Type{T}, x) = convert(T, x)
```
Then you can make this `_aesconvert.(Float64, aes.x)`.

trohwer:
There is an additional problem (related, but not quite the same) in the else branch
of if aes.color === nothing. I will send a pull request for that one.

bjarthur:
fixed by https://github.com/GiovineItalia/Gadfly.jl/pull/983 i believe instead

Issue #967, opened 2017-03-14
BoundsError with group aesthetic


jotwin:
works with color = :i  but not with group = :i

```
 plot(baddf, x=:phi, y=:fhat, group=:i, Geom.line)
Error showing value of type Gadfly.Plot:
ERROR: BoundsError: attempt to access 1-element Array{UInt32,1} at index [4]
 in permute!!(::Array{UInt32,1}, ::Array{Int64,1}) at ./combinatorics.jl:86
 in permute!! at /home/jakub/.julia/v0.5/DataArrays/src/pooleddataarray.jl:452 [inlined]
 in permute!(::DataArrays.PooledDataArray{ColorTypes.RGBA{Float32},UInt32,1}, ::Array{Int64,1}) at ./combinatorics.jl:107
 in render(::Gadfly.Geom.LineGeometry, ::Gadfly.Theme, ::Gadfly.Aesthetics) at /home/jakub/.julia/v0.5/Gadfly/src/geom/line.jl:112
 in render(::Gadfly.Geom.LineGeometry, ::Gadfly.Theme, ::Gadfly.Aesthetics, ::Array{Gadfly.Aesthetics,1}, ::Array{Gadfly.Data,1}, ::Dict{Symbol,Gadfly.ScaleElement}) at /home/jakub/.julia/v0.5/Gadfly/src/geometry.jl:48
 in (::Gadfly.##113#115{Dict{Symbol,Gadfly.ScaleElement}})(::Tuple{Gadfly.Layer,Gadfly.Aesthetics,Array{Gadfly.Aesthetics,1},Array{Gadfly.Data,1},Gadfly.Theme}) at ./<missing>:0
 in collect(::Base.Generator{Zip{Array{Gadfly.Layer,1},Zip{Array{Gadfly.Aesthetics,1},Zip{Array{Array{Gadfly.Aesthetics,1},1},Base.Zip2{Array{Array{Gadfly.Data,1},1},Array{Gadfly.Theme,1}}}}},Gadfly.##113#115{Dict{Symbol,Gadfly.ScaleElement}}}) at ./array.jl:307
[...]

PR #979, opened 2017-04-13
Compose.line called with wrong object type in render

trohwer:
In case

elseif length(aes.color) == 1 &&
            !(isa(aes.color, PooledDataArray) && length(levels(aes.color)) > 1)

the Compose.line is called with the wrong type (it expects array of array of tuple; compare else branch).

### DIFF
--- a/src/geom/line.jl
+++ b/src/geom/line.jl
@@ -159,7 +159,7 @@ function render(geom::LineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics)
             sort!(points, by=first)
         end
 
-        ctx = compose!(ctx, Compose.line(points,geom.tag),
+        ctx = compose!(ctx, Compose.line([points],geom.tag),
                        stroke(aes.color[1]),
                        strokedash(line_style),
                        svgclass("geometry"))


Mattriks:
There is no need to change this. For example:
```Julia
using Compose
points = [(x, y) for (x, y) in zip(1:10, 2+randn(10))]

# Either line(points) or line([points]) works 
p = plot(x=[0, 10], y=[0,4],
    Guide.annotation(compose(context(), line(points), stroke(colorant"black"))))
```

trohwer:
It does not work with DateTime for x, e.g. try it with

points=[(DateTime(2017,1,1),1),(DateTime(2017,1,2),2)]

and this happens, when you use DateTime in x components.

Maybe there is another way to fix it; I can't see, what magic does the conversion from "array of tuple" to "array of array of tuple".

Mattriks:
I concur, here is my example:
```Julia
t = Date(2017,1,4)+Dates.Day.(1:20)
# t = DateTime(2017,1,4)+Dates.Day.(1:20)
points = [(x, y) for (x, y) in zip(t, 2+randn(20))]

# With Date or DateTime objects line([points]) is needed
p = plot(x=[t[1],t[end]], y=[0,4],
    Guide.annotation(compose(context(), line([points]), stroke(colorant"black"))))
```

bjarthur:
there are [other](https://github.com/GiovineItalia/Gadfly.jl/issues/967) issues with Geom.line too.  would it be best to fix them all in one fell swoop?

Issue #918, opened 2016-10-14
Swapping color and shape for binary variables is not isomorphic


tpospisi:
I've run into a plotting error in which choosing my aesthetics changes which points are distinguished.  I have two binary variables (color and shape).  When I use the following code:

`plot(df, x = "x", y = "y", color = "color", shape = "shape")`

I get this figure which is correct (it agrees with constraints I know hold):

![correct](https://cloud.githubusercontent.com/assets/8901856/19371547/0a4b7764-9182-11e6-97e2-d0cb9c276aa6.png)

However, when I use this code:

`plot(df, x = "x", y = "y", shape = "color", color = "shape")`

I get this figure which is not correct:

![incorrect](https://cloud.githubusercontent.com/assets/8901856/19371551/0daef78c-9182-11e6-8b68-fee98b91af0c.png)

The difference is most obvious at the farthest right point.  In the correct graph it's a red circle like the rest of the upper cluster.  In the incorrect graph it's red and square while the upper cluster is blue and square.

This should not be the case.  The two graphs should be isomorphic since points which share the same characteristics in the first graph should share the same characteristics in the second.

Data to reproduce here: [example-data.txt](https://github.com/GiovineItalia/Gadfly.jl/files/528425/example-data.txt)

Mattriks:
This is a bit confusing.  Why did you not do:
```julia
plot(df, x="x", y="y", shape="shape", color="color") 
```
in the second line of code above?

tpospisi:
The naming of the variables is somewhat confusing.  What I've called "color" is class labels and "shape" indicates support vectors for an SVM solution.

I had initially wanted to have the second mapping: `plot(df, x="x", y="y", shape="support", color="class")` but as mentioned above, this doesn't make the correct figure.  On a whim I then decided to try the first mapping `plot(df, x="x", y="y", color="support", shape="class")` which does make the correct figure.

tlnagy:
Could you create a minimal example? It would help us diagnose the issue here because `color` and `shape` should be isomorphic.

tpospisi:
Here's a minimal example
```
x = [1, 2, 3, 4, 5, 6, 7, 8]
y = [0, 0, 0, 0, 0, 0, 0, 0]

a = [0, 0, 0, 0, 1, 1, 1, 1]
b = [1, 0, 1, 0, 1, 0, 1, 0]

plot(x = x, y = y, shape = a, color = b)
[...]

PR #991, opened 2017-04-25
WIP:  unify size,shape,color aesthetics for Geom.point

bjarthur:
still to do

- [x] time the refactor of render(Geom.point).  could be a lot slower
- [x] figure out why Coord.cartesian is necessary when size aesthetic is specified
- [x] think about a better way to `classify_data` the size aesthetic other than Union'ing it with `:functional`
- [x] more tests

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,7 +1,7 @@
 julia 0.5
 Colors 0.3.4
 Compat 0.18.0
-Compose 0.5
+Compose 0.5.1
 Contour 0.1.1
 DataFrames 0.4.2
 DataStructures
--- a/docs/src/lib/geoms/geom_point.md
+++ b/docs/src/lib/geoms/geom_point.md
@@ -10,7 +10,9 @@ The point geometry is used to draw various types of scatterplots.
 
   * `x`: X-axis position.
   * `y`: Y-axis position.
-  * `color` (optional): Point color (categorial or continuous).
+  * `color` (optional): Point color.  Categorical data will choose maximally distinguishable colors from the LCHab color space.  Continuous data will map onto LCHab as well.  Colors can also be specified explicitly for each data point with a vector of colors of length(x).  A vector of length one specifies the color to use for all points.  Default is Theme.default_color.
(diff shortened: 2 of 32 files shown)


Mattriks:
The size aesthetic currently works like so: `p = plot(x=1:5, y=randn(5), size=[x*mm for x in 1:5], Geom.point)
`. Are there new features anticipated here (apart from your 4 points above)?

bjarthur:
better to describe this PR as unifying how Geom.point treats the size, shape, and color aesthetics.  so for example, this now works `plot(x=1:5, y=randn(5), size=[x*mm for x in 1:5], color=rand([colorant"red",colorant"blue"], 5), Geom.point)`.

bjarthur:
can anyone explain why the Tick statistic takes into account the size aesthetic when setting axes limits?  i don't know of any geometry besides point which uses it.

Mattriks:
The answer to your q might be because size is a property of points, and points are used frequently in plotting - they are the main geometry (and probably the first geometry that a developer would write). 

I have a new q: will this PR fix #918?

tlnagy:
Oh yeah, #918 is nasty

bjarthur:
think this is ready to merge.  please send comments.

one thing to note is that the compiling the docs is throwing a world age error on julia 0.6.  works fine on julia 0.5.  as Gadfly's tests pass, i presume this is a problem with Documenter.jl.

tlnagy:
I'm going to try and give this a spin later today.


On Thu, Apr 27, 2017, at 11:11, Ben Arthur wrote:
> think this is ready to merge.  please send comments.
> 
> one thing to note is that the compiling the docs is throwing a world age
> error on julia 0.6.  works fine on julia 0.5.  as Gadfly's tests pass, i
> presume this is a problem with Documenter.jl.
> 
> -- 
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/GiovineItalia/Gadfly.jl/pull/991#issuecomment-297795083
---
Tamas Nagy
Graduate Student, Bioinformatics
Weiner Lab
University of California, San Francisco 
http://tamasnagy.com

tlnagy:
Ah shoot. I had totally forgotten about this. Everything looks good on my end.

bjarthur:
i figured no news was good news ;)

Issue #771, opened 2015-12-15
Geom.rectbin


jverzani:
It seems that these should be equivalent:

```
plot(x=[1],y=[1], Geom.rectbin)
```

and

```
plot(xmin=[1],ymin=[1], xmax=[1.5], ymax=[1.5], Geom.rectbin)
```

But the latter complains: The following aesthetics are required by RectbinStatistic but are not defined: y, x

Just commenting out `Gadfly.assert_aesthetics_defined("RectbinStatistic", aes, :x, :y)`  is not the solution.

Is there some other way to draw a rectangle with given dimensions?

protogeezer:
John - Nuts. This is something I’ve used. You are correct - commenting out the line raising the error won’t fix anything. It seems that the variable size rectangle support in Geom.rectbin got whacked in a commit back on 9/21.

A possible workaround:

http://stackoverflow.com/questions/23857793/julia-plotting-unknown-number-of-layers-in-gadfly

has worked in other contexts. Maybe put each rectangle in a separate layer, or all the rects of a single size in a layer. The latter would make it possible to have a deterministic number of layers.

Stephen

jverzani:
Thanks for the link. Hopefully this feature can be put back in, as adding primitives like a rectangle can be super useful. I'll take a look at the commit on 9/21 to see if it is something easy to do.

bjarthur:
fixed by https://github.com/GiovineItalia/Gadfly.jl/pull/993

Issue #845, opened 2016-05-16
Rectbin errors out when given max and min values


grero:
For some of my plots, I like to add a rectangular shaded region to highlight e.g. a sub-region used for a statistical comparison between two distributions. In Winston, I can accomplish this using the FillBetween type. I noticed that Gadfly has the ribbon geometry, but that seems to be bound to a line, i.e. it will only draw a ribbon around an existing line. It would be really useful to have a dedicated ShadedRegion (or a more fitting name) geometry for this use.

tbreloff:
You could just draw a polygon instead. Or make the line invisible (0 width
or transparent)

Mattriks:
According to the guide (http://dcjones.github.io/Gadfly.jl/geom_rectbin.html), `Geom.rectbin` is supposed to be able to do this for arbitrary rectangles, but I get an error when using the `xmin=, ymin=, xmax=, ymax=` option. So my solution (yet again!) is to make your own geometry:


```julia
using Colors, Compose, DataFrames

immutable RectangleGeometry <: Gadfly.GeometryElement 
     tag::Symbol 
 
    function RectangleGeometry(; tag::Symbol=Gadfly.Geom.empty_tag) 
         new(tag) 
     end 
end 

 
function Gadfly.element_aesthetics(::RectangleGeometry) 
    [:xmin, :ymin, :xmax, :ymax, :color] 
end 


function Gadfly.render(geom::RectangleGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics) 
 
     default_aes = Gadfly.Aesthetics() 
    default_aes.color = DataArrays.PooledDataArray(RGBA{Float32}[theme.default_color]) 
    aes = Gadfly.inherit(aes, default_aes)
    alpha_color = map(theme.lowlight_color, aes.color)
    
    ctx = context()
    compose!(ctx, rectangle(aes.xmin, aes.ymin, 
      aes.xmax-aes.xmin, aes.ymax-aes.ymin, geom.tag), fill(alpha_color)) 
    return compose!(context(order=4), svgclass("geometry"), ctx)
end

rect1 = RectangleGeometry

# in the next line, opacity = 0.2
theme1 = Theme(lowlight_color=c -> RGBA(c.r, c.g, c.b, 0.2))

D = DataFrame( x1=[-2,0], y1=[-1,0], x2=[1,4], y2=[1,2], col=["a","b"] )
p = plot(
[...]

PR #993, opened 2017-04-29
Geom.rect

Mattriks:
Two instances of `RectangularBinGeometry` are necessary:
- `Geom.rectbin` which calls `Stat.rectbin` (takes `x` & `y` aesthetics)
- `Geom.rect` which calls `Stat.Identity` (takes `xmin`, `ymin`, `xmax`, `ymax`)

The former currently exists, this PR adds the latter, and fixes issues #771 & #845.
Note the name `Geom.rect` is used rather than `Geom.rectangle` to avoid conflicts with `Compose.rectangle` (which is widely used without the package specifier in Gadfly code).

Here is an example:
```Julia
theme1 = Theme(default_color=RGBA(0, 0.75, 1.0, 0.5))
D = DataFrame(x=[0.5,1], y=[0.5,1], x1=[0,0.5], y1=[0,0.5], x2=[1,1.5], y2=[1,1.5])
pa = plot(D, x=:x, y=:y, Geom.rectbin, theme1)
pb = plot(D, xmin=:x1, ymin=:y1, xmax=:x2, ymax=:y2, Geom.rect, theme1)
hstack(pa, pb)
```
issue845a

### DIFF
--- a/docs/src/lib/geoms/geom_rectbin.md
+++ b/docs/src/lib/geoms/geom_rectbin.md
@@ -1,5 +1,5 @@
 ```@meta
-Author = "Daniel C. Jones"
+Author = "Daniel C. Jones, Mattriks"
 ```
 
 # Geom.rectbin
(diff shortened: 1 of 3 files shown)


bjarthur:
ggplot uses `geom_rect` and `geom_tile`.  should we follow suit?

http://ggplot2.tidyverse.org/reference/geom_tile.html

Mattriks:
I prefer `Geom.rectbin`, and I don't think we should always follow suit.

bjarthur:
i find `Geom.rectbin` confusing given that it doesn't actually bin the data, like `Geom.hexbin` does.

Mattriks:
My suggestion is to open a new issue about the naming of `Geom.rectbin` if you want to change it.

PR #999, opened 2017-05-17
Add DataArrays to REQUIRE

tkelman:
since it's used directly in src, it should be explicitly listed in REQUIRE
rather than assuming it will be present as a transitive dependency

(e.g. if there were to be some future version of DataFrames that did not depend on DataArrays)


bjarthur:
great catch.  i've submitted another [PR](https://github.com/GiovineItalia/Gadfly.jl/pull/1000) so that the tests pass for this one...

PR #1020, opened 2017-07-17
add Compose-style regression tests

bjarthur:
title says it all.  see the new section in the docs for an explanation of how it works.


bjarthur:
tests pass on 0.5 and 0.6.  please send your comments.  would like to merge soon.

tlnagy:
This seems like a much more robust system than what we had before. 👍

bjarthur:
is it safe to assume `Base.LibGit2` will work?  or that `LibGit2.jl` will work once it it's pushed out into a separate package?  if so, i can refactor the `run` calls to git to use it instead.  otherwise, how about just doing a `which git` and acting appropriately?

PR #1023, opened 2017-07-20
better support in {h,v,grid}stack for using context() as empty placeholder

bjarthur:
```vstack(plot, context, plot, context, ...)``` now works.  before the args had to either all be plots, or there could just be a pair with one of them being a context.


bjarthur:
also fixed the docs re. the argument order for `title`

Issue #816, opened 2016-03-31
Zooming in on a scatter plot enlarges the points


cstjean:
It used to be that zooming in (in IJulia) on a scatter plot would keep the visible dot size constant, but since ~January the behaviour is to grow the points as well:

![screen shot 2016-03-30 at 22 14 16](https://cloud.githubusercontent.com/assets/8622776/14163577/ccf7a4b2-f6c4-11e5-944f-b8a0304631b3.png)

This makes the zooming feature rather useless.

tlnagy:
That's pretty useless.

Issue #731, opened 2015-10-23
SVGJS backend - future features and current bugs


Godisemo:
I want this issue to collect a list of ideas for future features for the SVGJS backend to gadfly. Things like panning, zooming and toggling data. Please comment with your ideas and suggestions.

I also want to collect a list of known bugs related to the same backend. No core SVG rendering issues.

**Suggestions and ideas**
- [ ] Box zoom without aspect ratio limitations
- [ ] Crosshair which displays exact coordinate

**Known bugs**
- [ ] ...

gestay:
Zooming and panning are not working well on gadfly master.

Godisemo:
@gestay Could you elaborate?

gestay:
The zoom slide don't show, and the panning don't work.
I tried to record an animated gif of the behavior:

![gadfly](https://cloud.githubusercontent.com/assets/10811856/10777912/1df5238e-7cff-11e5-8885-6411ca0bd05b.gif)

gestay:
There are other cases where the zoom slide only appears when grid lines are touched.

![gadfly2](https://cloud.githubusercontent.com/assets/10811856/10778013/1d9c7c56-7d00-11e5-8c02-0de7e67ff11b.gif)

dcjones:
I can confirm this. There was a pretty big refactor of parts of Compose that just got merged, so I suspect I broke something with that. I'll fix this before I tag a new version.

dcjones:
@gestay Fixed in Compose master now.

gestay:
thanks!

gestay:
@Godisemo,  data toggling is not working for the following geometries:
* Geom.ribbon
* Geom.bar
* Geom.histogram
* Geom.polygon

tlnagy:
Also it would be nice to have a reset to default view option.

bjarthur:
+1 for box zoom w/o aspect ratio limitations.  @Godisemo do you have time to work on this?

Natharax:
Interactive toggling display of colors doesn't work for colors whose names contain periods/decimals.  This is evident in the gasoline example of the tutorial - drawing the plot to SVGJS allows all but U.K. and U.S.A. to be toggled.

bjarthur:
yep, i've noticed that too.  it's even more broken with https://github.com/GiovineItalia/Gadfly.jl/pull/1037.  will try to fix.  thx.

bjarthur:
[...]

PR #1037, opened 2017-08-29
enhanced interactivity

bjarthur:
fixes https://github.com/GiovineItalia/Gadfly.jl/issues/816, https://github.com/GiovineItalia/Gadfly.jl/issues/731

1. enable rectangular (instead of just square) zoom boxes
2. enable panning with keyboard
3. add crosshair to display coordinates
4. keep marker size fixed whatever the zoom level is
5. keep labels in same relative position no matter the zoom level

~~~the last is in parens because i need some help.  specifically, how does one convert into Gadfly coordinates the x,y coords which SnapSVG provides?  and i have yet to figure out a way to display them in the plot window instead of the console.~~~  WIP.

~~~also,~~~ the zoom slider in the upper right with the plus/minus buttons has been removed, because the zoom scale is now a 2-tuple, for the x and y dimensions, so a single slider didn't make sense.  please let me know if you have a problem with this as it is a major change in the GUI.


bjarthur:
@aviks @Mattriks @tlnagy @dcjones i think this is ready to merge.  as it's a big change to the pan/zoom API, could you please checkout this branch and give it a try?  you'll also need to checkout https://github.com/GiovineItalia/Compose.jl/pull/271.  thanks!

tlnagy:
I finally got a chance to check this out. The good:

- Saner behavior with point sizes when zoomed in
- Weird scaling issues fixed 

The bad:

- I don't see any crosshair or coordinates?
- Zoom not very discoverable without the slider
- Disorienting a bit. I think something like ImageJ's minimap could help a bunch:

image

I made a quick movie showing my interaction with the plot: https://youtu.be/pIMe8djbhhA Does this look like it should? I checked out this branch and the Compose one and plotted from the REPL using `using Gadfly; plot(x=rand(100), y=rand(100), Geom.point, Geom.smooth)`

bjarthur:
thanks for the comments.

the C key toggles on and off the display of the cursor coordinates in the upper right corner.

i agree it's not very discoverable, but it is clearly documented.

bjarthur:
this is ready to merge i think.  @tlnagy i've made an attempt to make the features more discoverable.  could you please try it out again?  you'll need to checkout this PR plus Compose master.  thx.

bjarthur:
the most recent commit in this PR makes the relative positions of labels invariant to zoom level.  give it a whirl!

also, re. adding an ImageJ-style minimap--  the ticks marks on the axes serve the same purpose, so i don't think that's needed here.  for an image, yes, but not for a plot.

tlnagy:
I see a `?` in the upper right corner? It's present even when I toggle the coordinates off. I checked out the latest Compose master and this PR

image

bjarthur:
the question mark is *not* present when the mouse is outside of the plot, correct?  and what happens when you press the question mark on the keyboard when the mouse is inside of the plot?

tlnagy:
Ah.

image

I still think it's a little janky looking. Would it be possible to use something like `awesome-question-circle-o` from here: https://leungwensen.github.io/svg-icon/#awesome. It's SVG and looks like its MIT licensed too.

tlnagy:
Great job btw, the interactivity is much improved.

Besides a couple aesthetic issues (the padding and color of the pop-up box) that we may disagree about, this PR is basically ready to merge.

bjarthur:
thanks.  i used `distinguishable_colors` against the background color to make sure it was visible against a dark theme, no matter what they chose.

the padding is a pain.  i already hack it by adding `1px`.  not sure how to make that robust to scaling.

i could try the question mark SVG graphic.  at this point though, i'm kinda tired of working on this PR!  that plus there's so many other things to be done.

what are your thoughts on merging this before or after the DataFrames 11 PR?  there's very little overlap so i don't think the merge will be hard either way.  i'm inclined to do so before since support for julia 0.5 will be dropped.  i'd also like to tag a release right before DF11 is merged so that there's one final version before we drop support for 0.5

tlnagy:
I'm okay with merging it as is. 

If I have some time, I'll see what I can do about the question mark. One PR doesn't have to do everything haha

bjarthur:
i'll spend a little more time on the padding (both on the cursor coordinates and the help box) and the question mark before merging.

tlnagy:
Okay. I've given a quick read-through and I'm okay with merging whenever you're satisfied with the tweaking. 👍

PR #1041, opened 2017-09-12
Misc. boxplot fixes and improvements

dcjones:
Fixes a bug and adds some options to control how boxplots are drawn.


bjarthur:
looks good to me.

PR #1045, opened 2017-10-09
Add point_label_font in themes.md

### LINKED ISSUE #1038: change the size of Geom.label
http://gadflyjl.org/stable/lib/geoms/geom_label.html#Geom.label-1

Can I change the size of text in Geom.plot?

I used these this
 theme(major_label_font_size=15pt, minor_label_font_size=15pt)

But it only applied to label in x or y axis. not applied to the label in the text in plot ( besides points)

Is there any other method?

jonathanBieler:
Fix #1038, I've also clarified a bit the minor_label_font one.


bjarthur:
thanks!

Issue #1019, opened 2017-07-14
Both Gadfly and Base export "cross"


YingboMa:
```
julia> using Gadfly

julia> cross
WARNING: both Gadfly and Base export "cross"; uses of it in module Main must be qualified
ERROR: UndefVarError: cross not defined
```

Is it possible to rename `Gadfly.cross` function?

bjarthur:
to what would you suggest?

YingboMa:
I actually have no idea. :laughing: Could you come up with one?

tlnagy:
Maybe call the Gadfly one `plussign`? The other option is to export all of Gadfly's shapes in a submodule so you'll have call Gadfly.Shape.Cross (like we do for Geoms or for Guides), which would prevent the conflict with Base.

ehdiix:
Is this issue still open?

bjarthur:
`plussign` is reasonable, but then should we rename `xcross` as well?

despite being more verbose, creating a `Shape` submodule is probably the right way to go.

PR #1050, opened 2017-11-06
gather circle, square, etc into new Shape module

bjarthur:
fixes https://github.com/GiovineItalia/Gadfly.jl/issues/1019

does anyone know how to deprecate `circle`, `square`, etc.?  they used to be exported, but now one must say `Shape.circle`, `Shape.square`, etc.  i tried `@deprecate` in various places, but it seems to be meant only for a name / argument change, and not for changes in whether it's exported or not.

also, since this is a breaking change, i believe we should increment the minor release version on the next tag.

lastly, is anyone else having trouble building the docs locally?  i'm getting errors:

```
$ julia docs/make.jl
Documenter: setting up build directory.
Documenter: expanding markdown templates.
ERROR: LoadError: UndefVarError: distinguishable_colors not defined
Stacktrace:
 [1] gen_colors(::Int64) at ./none:3
 [2] apply_scale(::Gadfly.Scale.DiscreteColorScale, ::Array{Gadfly.Aesthetics,1}, ::Gadfly.Data, ::Vararg{Gadfly.Data,N} where N) at /Users/arthurb/.julia/v0.6/Gadfly/src/scale.jl:502
```

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -22,6 +22,7 @@ makedocs(
             hide("Statistics" => "lib/stats.md", load_dir("stats")),
             hide("Coords" => "lib/coords.md", load_dir("coords")),
             hide("Scales" => "lib/scales.md", load_dir("scales")),
+            "Shapes" => "lib/shapes.md",
         ],
         "Development" => Any[
             "Rendering Pipeline" => "dev/pipeline.md",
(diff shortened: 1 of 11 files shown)


tlnagy:
Maybe ask on discourse.julialang.org? I don't actually know how you're supposed to do for changing export rules. 

> lastly, is anyone else having trouble building the docs locally? i'm getting errors:

I saw this too and traced this to `Colors.jl` not being imported explicitly. Maybe Gadfly used to re-export Colors and that behavior changed? Regardless, docs build with #1051 for me.

bjarthur:
thanks to @KristofferC for https://discourse.julialang.org/t/how-to-deprecate-a-change-in-exported-names/7257

Issue #1039, opened 2017-09-03
Test harness should not assume julia or git are on path, or prompt for input when run non-interactively


tkelman:
ref https://github.com/GiovineItalia/Gadfly.jl/pull/1024#commitcomment-24056139 and https://github.com/GiovineItalia/Gadfly.jl/pull/1020#commitcomment-24056160 - `Pkg.test("Gadfly")` shouldn't fail when git isn't installed, should be more careful about using the actually running julia instead of whatever happens to be on the path (which might not be any julia versions at all), and should never prompt when run programmatically. I wouldn't be surprised if the prompting makes Gadfly freeze and time out on PkgEval.

PR #1057, opened 2017-11-24
fix tests

bjarthur:
fixes https://github.com/GiovineItalia/Gadfly.jl/issues/1039

@tkelman this should fix your issue.  also, please remind me whether one should use `Pkg.dir` or `@__DIR__` inside of test/build scripts.


tkelman:
always `@__DIR__`. never use `Pkg.dir` in a package to refer to its own location, not guaranteed to be running from there

Issue #1018, opened 2017-07-11
Feature Request: Make Plot padding configurable


silvanheller:
Issue #654 brought this up, but was closed without discussion / reason. 
Currently, you can only pad the whole plot in all directions equally using _plot_padding_ -> [documentation](http://gadflyjl.org/stable/man/themes.html). It would be nice to have _xtop_, _xbottom_, _ytop_ and _ybottom_ options for plot padding. This is especially the case when working with String labels which can be cut off if they are on the last tick.

bjarthur:
i'd be happy to review a pull request.  at first glance it seems straightforward enough for someone new to Gadfly's internals to tackle.  would just need to add a few fields to `Theme` [here](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/theme.jl#L96), and then use them when rendering [here](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/Gadfly.jl#L758).

bjarthur:
fixed with https://github.com/GiovineItalia/Gadfly.jl/pull/1081#issuecomment-355653730

PR #1078, opened 2017-12-30
Configure plot_padding

Mattriks:
Here is my two line solution to the `plot_padding` issue (#1018). Plus I updated the `Theme` doc file. Should I add a test file?  Some examples:

```Julia
 using Compose, DataFrames, Gadfly
 Gadfly.push_theme(Theme(default_color=colorant"black", background_color="white"))

# Note the padding order is fixed by Compose.jl:
# plot_padding = [left, right, top, bottom]

xscale = Scale.x_continuous(format=:plain)
gp = Geom.polygon(preserve_order=true)
D = DataFrame(x=[0, 0, 10^6, 10^6], y=[0, 10, 10, 0])

# Padding
# In absolute units:
pa = plot(D, x=:x, y=:y, gp, xscale, style(plot_padding=[10,10,5,5].*mm))
# In relative units (relative to width & height of plot):
pb = plot(D, x=:x, y=:y, gp, xscale, style(plot_padding=[0.05w,0.05w,0.2h,0.2h]))
# 1mm padding on all sides, note x-axis right label gets cut deliberately:
pc = plot(D, x=:x, y=:y, gp, xscale, style(plot_padding=[1mm]))
hstack(pa,pb,pc)

```
issue1018

### DIFF
--- a/docs/src/man/themes.md
+++ b/docs/src/man/themes.md
@@ -39,7 +39,7 @@ These parameters can either be used with `Theme` or `style`
     Nothing)
   * `background_color`: Background color for the entire plot. If nothing, no
     background. (Color or Nothing)
-  * `plot_padding`: How much padding should be put around the plot as a whole (Measure)
+  * `plot_padding`: Padding around the plot. The order of padding is: `plot_padding=[left, right, top, bottom]`. If a vector of length one is provided e.g.  `[5mm]` then that value is applied to all sides. Absolute or relative units can be used. (Vector{<:Measure})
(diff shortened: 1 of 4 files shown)


bjarthur:
CI failing on julia 0.5.  i expect will drop support for 0.5 when 0.7 comes out.  but until then we should support 0.5.  can you fix it?  otherwise this looks great.

the example you provide above is the test.  all you have to do is add it to testscripts/.

thanks for this!

bjarthur:
thanks for this!

Issue #1083, opened 2018-01-02
No method matching isfinite for nullable Int64


alexomics:
```
julia> versioninfo()
Julia Version 0.6.2
Commit d386e40c17 (2017-12-13 18:08 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i5-2415M CPU @ 2.30GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, sandybridge)

julia> Pkg.status("Gadfly")
 - Gadfly                        0.6.4

julia> head(s)
6×2 DataFrames.DataFrame
│ Row │ foo │ bar     │
├─────┼─────┼─────────┤
│ 1   │ 521 │ 40.0    │
│ 2   │ 521 │ 40.0003 │
│ 3   │ 542 │ 40.0005 │
│ 4   │ 540 │ 40.0007 │
│ 5   │ 566 │ 40.001  │
│ 6   │ 664 │ 40.0012 │

julia> plot(s, x="bar", y="foo", Geom.line)
Error showing value of type Gadfly.Plot:
ERROR: MethodError: no method matching isfinite(::Nullable{Int64})
Closest candidates are:
  isfinite(::Float16) at float.jl:539
  isfinite(::BigFloat) at mpfr.jl:831
  isfinite(::DataArrays.NAtype) at /Users/Alex/.julia/v0.6/DataArrays/src/predicates.jl:10
  ...
Stacktrace:
 [1] apply_statistic_typed(::Nullable{Int64}, ::Nullable{Int64}, ::Array{Nullable{Int64},1}, ::Array{Void,1}, ::Array{Void,1}) at /Users/Alex/.julia/v0.6/Gadfly/src/statistics.jl:909
 [2] apply_statistic(::Gadfly.Stat.TickStatistic, ::Dict{Symbol,Gadfly.ScaleElement}, ::Gadfly.Coord.Cartesian, ::Gadfly.Aesthetics) at /Users/Alex/.julia/v0.6/Gadfly/src/statistics.jl:772
[...]

Issue #1065, opened 2017-12-06
Dataframes v0.11 is out


tlnagy:
This is a pretty big change and we need make sure Gadfly works with it. I believe the main hangup right now is that RDatasets (which we use for tests) doesn't work with v0.11 until https://github.com/johnmyleswhite/RDatasets.jl/pull/48 merges.

Then we'll need to drop `DataArrays` from `REQUIRE` and peg the minimum version of `DataFrames` at v0.11 and fix any problems.

va-andrew:
Plus one for this issue. We had to remove Gadfly from our workflow until it can coexist with the newer versions of `DataFrames` (it currently forces a downgrade when used with `Pkg`).

tlnagy:
Looks like https://github.com/johnmyleswhite/RDatasets.jl/pull/48 is basically ready so we can bump our requirements once it's merged and a new release is tagged.

tlnagy:
The work has moved to #1090

andreasnoack:
I think this can be closed. Eventually, we might want to remove `DataArrays` as a dependency but it is not required.

bjarthur:
happy to close this.  you've done great work @andreasnoack !  thanks for taking on this important issue.

re. `DataArrays`...  is it or is it not a dependency?  should be removed now from REQUIRE if not.

andreasnoack:
There is still code left to handle `DataArrays` if passed to Gadfly but I don't think we use them for anything so I suspect that it can be removed. I might take a look at that later today.

PR #1088, opened 2018-01-11
[WIP] work towards updating to DataFrames v0.11+ framework

tlnagy:
i haven't been able to fix everything due to slight inconsistencies between PooledDataArrays and CategoricalArrays behavior. See https://github.com/JuliaData/CategoricalArrays.jl/issues/117 for
details. Working towards a fix for https://github.com/GiovineItalia/Gadfly.jl/issues/1065.

It would nice to have this soon as we're holding back the whole DataFrames ecosystem until we do. Maybe @alyst and @nalimilan could also provide us some feedback to get this working with v0.11

### DIFF
--- a/src/aesthetics.jl
+++ b/src/aesthetics.jl
@@ -1,8 +1,8 @@
 const NumericalOrCategoricalAesthetic =
-    Union{(Void), Vector, DataArray, PooledDataArray}
+    Union{(Void), Vector, DataArray, CategoricalArray}
 
 const CategoricalAesthetic =
-    Union{(Void), PooledDataArray}
+    Union{(Void), CategoricalArray}
 
 const NumericalAesthetic =
     Union{(Void), Matrix, Vector, DataArray}
@@ -287,18 +287,8 @@ function cat_aes_var!{T, U}(a::AbstractArray{T}, b::AbstractArray{U})
     return ab
 end
 
-function cat_aes_var!{T}(xs::PooledDataVector{T}, ys::PooledDataVector{T})
-    newpool = T[x for x in union(Set(xs.pool), Set(ys.pool))]
-    newdata = vcat(T[x for x in xs], T[y for y in ys])
-    PooledDataArray(newdata, newpool, [false for _ in newdata])
-end
-
-function cat_aes_var!{T, U}(xs::PooledDataVector{T}, ys::PooledDataVector{U})
-    V = promote_type(T, U)
-    newpool = V[x for x in union(Set(xs.pool), Set(ys.pool))]
-    newdata = vcat(V[x for x in xs], V[y for y in ys])
-    PooledDataArray(newdata, newpool, [false for _ in newdata])
-end
+cat_aes_var!{T}(xs::CategoricalArray{T}, ys::CategoricalArray{T}) = vcat(xs, ys)
+cat_aes_var!{T, U}(xs::CategoricalArray{T}, ys::CategoricalArray{U}) = vcat(xs, ys)
 
 
 # Summarizing aesthetics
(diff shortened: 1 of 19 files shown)


bjarthur:
thanks for tackling this.  i don't use DataFrames much, so am not familiar with the API changes.  please though make use of the [regression testing](http://gadflyjl.org/latest/dev/regression.html) infrastructure in Gadfly before you merge.

tlnagy:
I definitely don't plan on merging this until it's thoroughly vetted since it's a lot of major changes.

src/misc.jl
```diff
@@ -1,13 +1,13 @@
 #Is this usable data?
-isconcrete{T<:Number}(x::T) = !isna(x) && isfinite(x)
+isconcrete{T<:Number}(x::T) = !ismissing(x) && isfinite(x)
```
alyst: I guess it should be something like:
```julia
isconcrete(::Missing) = false
isconcrete(x::Number) = isfinite(x)
```

tlnagy: Good catch. This should fix https://github.com/GiovineItalia/Gadfly.jl/issues/1083 too, I think

nalimilan:
Cc: @andreasnoack

andreasnoack:
Doh. I've been through almost all the same changes earlier this week. A good reminder to open a `WIP` PR early to communicate the efforts and avoid wasted efforts.

I had a lot of issues with the wrapping behavior of `CategoricalArrays` so I actually tried to look into using https://github.com/JuliaArrays/IndirectArrays.jl internally in `Gadfly` while, of course, still allowing `CategoricalArray`s input. In particular, I think that `Compose` should remain agnostic to `CategoricalArray`s but it is was a bit challenging to avoid that `Gadfly` sends off `CategoricalArray`s or `CategoricalValue`s to `Compose`.

tlnagy:
> I've been through almost all the same changes earlier this week.

It would be great if you could push your changes to this PR or make a new PR, which ever is easier.

> In particular, I think that `Compose` should remain agnostic to `CategoricalArrays`

Agreed. And this was where I got hung up as well.

Mattriks:
Out of interest, JuliaArrays/IndirectArrays.jl#12

src/aesthetics.jl
```diff
@@ -287,18 +287,8 @@ function cat_aes_var!{T, U}(a::AbstractArray{T}, b::AbstractArray{U})
...
-end
-
-function cat_aes_var!{T, U}(xs::PooledDataVector{T}, ys::PooledDataVector{U})
-    V = promote_type(T, U)
-    newpool = V[x for x in union(Set(xs.pool), Set(ys.pool))]
-    newdata = vcat(V[x for x in xs], V[y for y in ys])
-    PooledDataArray(newdata, newpool, [false for _ in newdata])
-end
+cat_aes_var!{T}(xs::CategoricalArray{T}, ys::CategoricalArray{T}) = vcat(xs, ys)
+cat_aes_var!{T, U}(xs::CategoricalArray{T}, ys::CategoricalArray{U}) = vcat(xs, ys)
```
alyst: BTW, `cat_aes_var{T}(...)` is pre-0.6 syntax. Since DataFrames 0.11 requires 0.6, it's better write it as `cat_aes_var(...) where {T}`, although in this particular case it could be as simple as (replaces both definitions)
```julia
cat_aes_var!(xs::CategoricalArray, ys::CategoricalArray) = vcat(xs, ys)
```

src/aesthetics.jl
```diff
@@ -366,9 +351,8 @@ function by_xy_group{T <: Union{Data, Aesthetics}}(aes::T, xgroup, ygroup,
             end
 
             for i in 1:n, j in 1:m
-                if typeof(vals) <: PooledDataArray
-                    setfield!(aes_grid[i, j], var,
-                              make_pooled_data_array(typeof(vals), staging[i, j]))
+                if typeof(vals) <: CategoricalArray
```
alyst: `vals isa CategoricalArray` or `isa(vals, CategoricalArray)` is shorter

tlnagy:
Closed in favor of https://github.com/GiovineItalia/Gadfly.jl/pull/1090

Issue #1007, opened 2017-05-22
alter travis to *not* checkout Compose


bjarthur:
the travis yml currently checks out master on Compose before testing Gadfly.  this has recently caused a couple of problems to be experienced by users that where not caught by the automated tests.  i'd suggest removing it, but am curious why it was put there in the first place.

tlnagy:
I think it was adopted when Compose was split out of Gadfly. The two projects always moved in lockstep and as long as changes to Compose are tagged whenever changes in Gadfly are tagged it's okay. Basically, I think it was done for convenience.

bjarthur:
the problem arose precisely when i didn't think to tag Compose when tagging Gadfly.

it seems to me that developers will have master checked out on both, and so will know when testing locally whether the master branches are compatible.  i'm hence inclined to think that travis should *not* checkout master on Compose, so that the automated testing makes sure it works for users who don't have master checked out.

i hesitate though because even that is not fool proof:  if Compose is tagged, there is no automated test to make sure that the tagged or even master branch of Gadfly works with it.  not sure what to do about that case other than to try and remember to test Gadfly locally when tagging Compose.

aviks:
Given the relationship between Gadfly and Compose, this is the best possible situation right now. We just have to remember to tag Compose and Gadfly together, always. They should have the same version number always, that makes things easier to track and trace. 

Longer term, if and when the API of Compose stabilises, we can do something different. But right now, leaving things as is seems to be the path of least resistance.

PR #1090, opened 2018-01-22
DataFrames 0.11+ compat via IndirectArrays

andreasnoack:
@tlnagy Here is my version of the DataFrame 0.11+ compatible version. This version passes locally (see caveat below) with [this version of IndirectArrays](https://github.com/JuliaArrays/IndirectArrays.jl/pull/12). I decided that the wrapping behavior of CategoricalArrays was too inconvenient to get working with the current setup because it is implicitly assumed in most functions that "categorical" arrays return the same kand of values as non-categorical arrays. Maybe by restructuring the code to rely more on dispatch to smaller specialized methods it might be possible to use CategoricalArrays internally, but I'm not sure it is worth it. Basically, I don't think it is useful to have the `pool` handy for the use cases here in Gadfly.

Although the tests pass, it requires that Julia runs with `--compilecache=no`. Otherwise, it triggers what appears to be completely spurious ambiguity errors that we'd need @vtjnash or @JeffBezanson to comment on. It is hard for me to narrow down the issue but it could be related to `module`/`using`/`include`. E.g. I can trigger/detrigger the issue by including a file where all lines are commented. I've also been able to trigger/detrigger it with a `foo() = 1` definition in the middle of the `statistics.jl` file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,7 @@ script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia -e 'Pkg.clone(pwd()); Pkg.build("Gadfly");'
   - julia -e 'Pkg.checkout("Compose")'
-  # because of a bug in Julia 0.6.2 we temporarily use --compilecache=no.
-  # The bug is fixed on master and will be included in 0.6.3 so when it
-  # has been releases, we can bump the requirement and remove the flag.
-  - julia --check-bounds=yes --compilecache=no -e 'Pkg.test("Gadfly", coverage=true)'
+  - julia --check-bounds=yes -e 'Pkg.test("Gadfly", coverage=true)'
 after_success:
   - julia -e 'cd(Pkg.dir("Gadfly")); Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
   - julia -e 'Pkg.checkout("Compose")'
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,17 @@ matrix:
     - julia: nightly
 notifications:
   email: false
+matrix:
+  allow_failures:
+  - julia: nightly
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia -e 'Pkg.clone(pwd()); Pkg.build("Gadfly");'
   - julia -e 'Pkg.checkout("Compose")'
-  - julia --check-bounds=yes -e 'Pkg.test("Gadfly", coverage=true)'
+  # because of a bug in Julia 0.6.2 we temporarily use --compilecache=no.
+  # The bug is fixed on master and will be included in 0.6.3 so when it
+  # has been releases, we can bump the requirement and remove the flag.
+  - julia --check-bounds=yes --compilecache=no -e 'Pkg.test("Gadfly", coverage=true)'
 after_success:
   - julia -e 'cd(Pkg.dir("Gadfly")); Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
   - julia -e 'Pkg.checkout("Compose")'
--- a/REQUIRE
+++ b/REQUIRE
@@ -16,5 +16,10 @@ Loess
 Showoff 0.0.3
 StatsBase
 Juno
-IndirectArrays 0.4
+IndirectArrays 0.4.1
 Missings
+# Gadfly doesn't use WeakRefString directly but because of a but in Julia 0.6.2, Gadfly
+# will trigger a ambiguity error between WeakRefString methods. A temporary method was
+# added in version WeakRefStrings 0.4.3 to work around this. When Gadfly requires at least
+# Julia 0.6.3, this dependency can therefore be dropped.
+WeakRefStrings 0.4.3
--- a/REQUIRE
+++ b/REQUIRE
@@ -16,5 +16,5 @@ Loess
 Showoff 0.0.3
 StatsBase
 Juno
-IndirectArrays
+IndirectArrays 0.4
 Missings
--- a/REQUIRE
+++ b/REQUIRE
@@ -18,7 +18,7 @@ StatsBase
 Juno
 IndirectArrays 0.4.1
 Missings
-# Gadfly doesn't use WeakRefString directly but because of a but in Julia 0.6.2, Gadfly
+# Gadfly doesn't use WeakRefString directly but because of a bug in Julia 0.6.2, Gadfly
 # will trigger a ambiguity error between WeakRefString methods. A temporary method was
 # added in version WeakRefStrings 0.4.3 to work around this. When Gadfly requires at least
 # Julia 0.6.3, this dependency can therefore be dropped.
--- a/REQUIRE
+++ b/REQUIRE
@@ -16,5 +16,5 @@ Loess
 Showoff 0.0.3
 StatsBase
 Juno
-PooledArrays
+IndirectArrays
 Missings
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,10 +1,10 @@
-julia 0.5
+julia 0.6
 Colors 0.3.4
 Compat 0.18.0
 Compose 0.5.2
 Contour 0.1.1
 CoupledFields
-DataFrames 0.4.2 0.11.0
+DataFrames 0.11.4
 DataArrays
 DataStructures
 Distributions
@@ -16,3 +16,5 @@ Loess
 Showoff 0.0.3
 StatsBase
 Juno
+PooledArrays
+Missings
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -10,7 +10,7 @@ using DataFrames
 using DataStructures
 using JSON
 using Showoff
-using PooledArrays
+using IndirectArrays
 using CategoricalArrays
 
 import IterTools
@@ -37,17 +37,6 @@ export Plot, Layer, Theme, Col, Row, Scale, Coord, Geom, Guide, Stat, Shape, ren
 @deprecate octogon Shape.octogon
 @deprecate hline Shape.hline
 @deprecate vline Shape.vline
-
-# Things that should go into other packages.
-# (Probably in a more efficient version)
-## PooledArrays
-function Base.append!(x::PooledArray{T,R,N,RA}, y::PooledArray{T,R,N,RA}) where {T,R,N,RA}
-    if x.pool == y.pool
-        append!(x.refs, y.refs)
-        return x
-    end
-    throw(ArgumentError("pools not identical"))
-end
 
 # Re-export some essentials from Compose
 export SVGJS, SVG, PGF, PNG, PS, PDF, draw, inch, mm, cm, px, pt, color, @colorant_str, vstack, hstack, title, gridstack
@@ -1126,7 +1115,7 @@ end
 
 import Juno: Juno, @render, media, Media
 
-media(Plot, Media.Plot)
+# media(Plot, Media.Plot)
 
 @render Juno.PlotPane p::Plot begin
     x, y = Juno.plotsize()
@@ -1232,9 +1221,9 @@ end
 
 # Determine whether the input is categorical or numerical
 
-const CategoricalType = Union{AbstractString, Bool, Symbol}
+const CategoricalType = Union{AbstractString, Bool, Symbol, CategoricalArrays.CategoricalValue, CategoricalArrays.CategoricalString}
 
-classify_data{N, T <: CategoricalType}(data::AbstractArray{T, N}) = :categorical
+classify_data{N, T <: Union{CategoricalType,Missing}}(data::AbstractArray{T, N})        = :categorical
 classify_data{N, T <: Union{Base.Callable,Measure,Colorant}}(data::AbstractArray{T, N}) = :functional
 classify_data{T <: Base.Callable}(data::T) = :functional
 classify_data(data::AbstractArray) = :numerical
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -11,6 +11,7 @@ using DataStructures
 using JSON
 using Showoff
 using PooledArrays
+using CategoricalArrays
 
 import IterTools
 import IterTools: distinct, drop, chain
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -1221,10 +1221,11 @@ end
 
 # Determine whether the input is categorical or numerical
 
-const CategoricalType = Union{AbstractString, Bool, Symbol, CategoricalArrays.CategoricalValue, CategoricalArrays.CategoricalString}
+const CategoricalType = Union{AbstractString, Bool, Symbol}
 
 classify_data{N, T <: Union{CategoricalType,Missing}}(data::AbstractArray{T, N})        = :categorical
 classify_data{N, T <: Union{Base.Callable,Measure,Colorant}}(data::AbstractArray{T, N}) = :functional
+classify_data(data::CategoricalArray) = :categorical
 classify_data{T <: Base.Callable}(data::T) = :functional
 classify_data(data::AbstractArray) = :numerical
 classify_data(data::Distribution) = :distribution
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -1115,7 +1115,7 @@ end
 
 import Juno: Juno, @render, media, Media
 
-# media(Plot, Media.Plot)
+media(Plot, Media.Plot)
 
 @render Juno.PlotPane p::Plot begin
     x, y = Juno.plotsize()
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -10,6 +10,7 @@ using DataFrames
 using DataStructures
 using JSON
 using Showoff
+using PooledArrays
 
 import IterTools
 import IterTools: distinct, drop, chain
@@ -35,6 +36,17 @@ export Plot, Layer, Theme, Col, Row, Scale, Coord, Geom, Guide, Stat, Shape, ren
 @deprecate octogon Shape.octogon
 @deprecate hline Shape.hline
 @deprecate vline Shape.vline
+
+# Things that should go into other packages.
+# (Probably in a more efficient version)
+## PooledArrays
+function Base.append!(x::PooledArray{T,R,N,RA}, y::PooledArray{T,R,N,RA}) where {T,R,N,RA}
+    if x.pool == y.pool
+        append!(x.refs, y.refs)
+        return x
+    end
+    throw(ArgumentError("pools not identical"))
+end
 
 # Re-export some essentials from Compose
 export SVGJS, SVG, PGF, PNG, PS, PDF, draw, inch, mm, cm, px, pt, color, @colorant_str, vstack, hstack, title, gridstack
--- a/src/geom/line.jl
+++ b/src/geom/line.jl
@@ -58,7 +58,7 @@ function render(geom::LineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics)
                                           element_aesthetics(geom)...)
 
     default_aes = Gadfly.Aesthetics()
-    default_aes.color = PooledArray(RGBA{Float32}[theme.default_color])
+    default_aes.color = IndirectArray(RGBA{Float32}[theme.default_color])
     aes = inherit(aes, default_aes)
 
     ctx = context(order=geom.order)
@@ -132,7 +132,7 @@ function render(geom::LineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics)
                       svgclass("geometry"))
 
     elseif length(aes.color) == 1 &&
-            !(isa(aes.color, PooledArray) && length(levels(aes.color)) > 1)
+            !(isa(aes.color, IndirectArray) && length(levels(aes.color)) > 1)
         T = (Tuple{eltype(aes.x), eltype(aes.y)})
         points = T[(x, y) for (x, y) in zip(aes.x, aes.y)]
         geom.preserve_order || sort!(points, by=first)
--- a/src/geom/line.jl
+++ b/src/geom/line.jl
@@ -58,7 +58,7 @@ function render(geom::LineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics)
                                           element_aesthetics(geom)...)
 
     default_aes = Gadfly.Aesthetics()
-    default_aes.color = PooledDataArray(RGBA{Float32}[theme.default_color])
+    default_aes.color = PooledArray(RGBA{Float32}[theme.default_color])
     aes = inherit(aes, default_aes)
 
     ctx = context(order=geom.order)
@@ -132,7 +132,7 @@ function render(geom::LineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics)
                       svgclass("geometry"))
 
     elseif length(aes.color) == 1 &&
-            !(isa(aes.color, PooledDataArray) && length(levels(aes.color)) > 1)
+            !(isa(aes.color, PooledArray) && length(levels(aes.color)) > 1)
         T = (Tuple{eltype(aes.x), eltype(aes.y)})
         points = T[(x, y) for (x, y) in zip(aes.x, aes.y)]
         geom.preserve_order || sort!(points, by=first)
--- a/src/geom/line.jl
+++ b/src/geom/line.jl
@@ -132,7 +132,7 @@ function render(geom::LineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics)
                       svgclass("geometry"))
 
     elseif length(aes.color) == 1 &&
-            !(isa(aes.color, IndirectArray) && length(filter(!ismissing, aes.color.values)) > 1)
+            !(isa(aes.color, IndirectArray) && count(!ismissing, aes.color.values) > 1)
         T = (Tuple{eltype(aes.x), eltype(aes.y)})
         points = T[(x, y) for (x, y) in zip(aes.x, aes.y)]
         geom.preserve_order || sort!(points, by=first)
--- a/src/geom/line.jl
+++ b/src/geom/line.jl
@@ -132,7 +132,7 @@ function render(geom::LineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics)
                       svgclass("geometry"))
 
     elseif length(aes.color) == 1 &&
-            !(isa(aes.color, IndirectArray) && length(levels(aes.color)) > 1)
+            !(isa(aes.color, IndirectArray) && length(filter(!ismissing, aes.color.values)) > 1)
         T = (Tuple{eltype(aes.x), eltype(aes.y)})
         points = T[(x, y) for (x, y) in zip(aes.x, aes.y)]
         geom.preserve_order || sort!(points, by=first)
--- a/src/scale.jl
+++ b/src/scale.jl
@@ -495,7 +495,9 @@ function apply_scale(scale::DiscreteColorScale,
     for (aes, data) in zip(aess, datas)
         data.color === nothing && continue
         for d in data.color
-            push!(levelset, d)
+            # Remove missing values
+            # FixMe! The handling of missing values shouldn't be this scattered across the source
+            ismissing(d) || push!(levelset, d)
         end
     end
 
@@ -514,15 +516,11 @@ function apply_scale(scale::DiscreteColorScale,
 
     for (aes, data) in zip(aess, datas)
         data.color === nothing && continue
-        ds = discretize(data.color, scale_levels)
-        colorvals = Array{RGB{Float32}}(count(!iszero, ds.index))
-        i = 1
-        for k in ds.index
-            if k != 0
-                colorvals[i] = colors[k]
-                i += 1
-            end
-        end
+        # Remove missing values
+        # FixMe! The handling of missing values shouldn't be this scattered across the source
+        ds = discretize([c for c in data.color if !ismissing(c)], scale_levels)
+
+        colorvals = colors[ds.index]
 
         colored_ds = IndirectArray(colorvals, colors)
         aes.color = colored_ds
--- a/src/scale.jl
+++ b/src/scale.jl
@@ -7,6 +7,7 @@ using DataArrays
 using DataStructures
 using Gadfly
 using Showoff
+using PooledArrays
 
 import Gadfly: element_aesthetics, isconcrete, concrete_length,
                nonzero_length
@@ -202,7 +203,7 @@ function apply_scale(scale::ContinuousScale,
                 T = Measure
             end
 
-            ds = Gadfly.hasna(vals) ? DataArray(T, length(vals)) : Array{T}(length(vals))
+            ds = any(ismissing, vals) ? DataArray(T, length(vals)) : Array{T}(length(vals))
             apply_scale_typed!(ds, vals, scale)
 
             if var == :xviewmin || var == :xviewmax ||
@@ -251,43 +252,43 @@ function apply_scale_typed!(ds, field, scale::ContinuousScale)
 end
 
 # Reorder the levels of a pooled data array
-function reorder_levels(da::PooledDataArray, order::AbstractVector)
+function reorder_levels(da::PooledArray, order::AbstractVector)
     level_values = levels(da)
     length(order) != length(level_values) &&
             error("Discrete scale order is not of the same length as the data's levels.")
     permute!(level_values, order)
-    return PooledDataArray(da, level_values)
+    return PooledArray(da, level_values)
 end
 
 function discretize_make_pda(values::Vector, levels=nothing)
     if levels == nothing
-        return PooledDataArray(values)
+        return PooledArray(values)
     else
-        return PooledDataArray(convert(Vector{eltype(levels)}, values), levels)
+        return PooledArray(convert(Vector{eltype(levels)}, values), levels)
     end
 end
 
 function discretize_make_pda(values::DataArray, levels=nothing)
     if levels == nothing
-        return PooledDataArray(values)
+        return PooledArray(values)
     else
-        return PooledDataArray(convert(DataArray{eltype(levels)}, values), levels)
+        return PooledArray(convert(DataArray{eltype(levels)}, values), levels)
     end
 end
 
 function discretize_make_pda(values::Range, levels=nothing)
     if levels == nothing
-        return PooledDataArray(collect(values))
+        return PooledArray(collect(values))
     else
-        return PooledDataArray(collect(values), levels)
+        return PooledArray(collect(values), levels)
     end
 end
 
-function discretize_make_pda(values::PooledDataArray, levels=nothing)
+function discretize_make_pda(values::PooledArray, levels=nothing)
     if levels == nothing
         return values
     else
-        return PooledDataArray(values, convert(Vector{eltype(values)}, levels))
+        return PooledArray(values, convert(Vector{eltype(values)}, levels))
     end
 end
 
@@ -329,7 +330,7 @@ immutable DiscreteScale <: Gadfly.ScaleElement
     labels::Union{(Void), Function}
 
     # If non-nothing, give values for the scale. Order will be respected and
-    # anything in the data that's not represented in values will be set to NA.
+    # anything in the data that's not represented in values will be set to missing.
     levels::Union{(Void), AbstractVector}
 
     # If non-nothing, a permutation of the pool of values.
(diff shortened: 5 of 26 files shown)


src/Gadfly.jl
```diff
@@ -1121,7 +1123,7 @@ end
 
 import Juno: Juno, @render, media, Media, Hiccup
 
-media(Plot, Media.Plot)
+# media(Plot, Media.Plot)
```
pfitzseb: This will break Juno's plot rendering.

andreasnoack:
So the issue mentioned above is fixed by https://github.com/JuliaLang/julia/pull/25832. I've tried to cherry-pick the fix on `release-0.6` branch and it works. This means that getting Gadfly working with the new Missing based data stack requires a new 0.6.x release which currently is not on the roadmap because of resource constraints. Hence there is a risk that we'll only get updated Gadfly supported after 0.7/1.0 is out. As a consequence, I've started the process of getting this branch working on Julia master.

tlnagy:
> Hence there is a risk that we'll only get updated Gadfly supported after 0.7/1.0 is out.

I think we're getting pretty close to a 0.7 release, no? I think that at this point it's been several months since the new DataFrames came out and the damage is likely done (people either pinned DataFrames or dropped Gadfly) so I wouldn't mind waiting till 0.7 came out. 

> As a consequence, I've started the process of getting this branch working on Julia master.

Awesome!

bjarthur:
@andreasnoack could you add `--compilecache=no` to this PR in travis.yml?  if it passes with that i'd suggest we merge.  could also add a temporary warning to the README/docs about needing to run with that flag.

andreasnoack:
I'd need to finish up https://github.com/JuliaArrays/IndirectArrays.jl/pull/12 first but then I might try to add the flag here.

bjarthur:
i tried checking out this PR, the PR in IndirectArrays, and using `--compilecache=no`.  works great except `test/testscripts/boxplot.jl`.  throws a `DimensionMismatch` error.  can you confirm?

andreasnoack:
I'm able to reproduce that error. I thought I had them all working. Will take a look.

andreasnoack:
Oh. Now I know what is going on. I had added what I thought was a harmless deprecation fix. Should work now.

Btw. There'll be some issues with the testing when we transition to 0.7 because 0.7 doesn't produce the same random `Int64`s as 0.7 for the same feeds. Hence the uuids embedded in the produces images will be different so we'd need a way to handle that in the regression testing.

bjarthur:
can we use a different seed to reproduce the old behavior?

andreasnoack:
Unfortunately, I don't think that is possible. See also https://github.com/JuliaLang/julia/pull/25058#issuecomment-362706966

kleinschmidt:
It looks like there's an issue somewhere here with levels/sorting.  I'm getting a mismatch between group labels and the actual data plotted using `Geom.subplot_grid`.

src/scale.jl
```diff
@@ -251,45 +253,51 @@ function apply_scale_typed!(ds, field, scale::ContinuousScale)
...
+discretize_make_pda(values::IndirectArray)         = values
+discretize_make_pda(values::IndirectArray, ::Void) = values
+discretize_make_pda(values::IndirectArray, levels) =
+    IndirectArray(values, convert(Vector{eltype(values)}, levels))
+
+discretize_make_pda(values::CategoricalArray)         = discretize_make_pda(values, unique(values))
+discretize_make_pda(values::CategoricalArray, ::Void) = discretize_make_pda(values)
+function discretize_make_pda(values::CategoricalArray{T}, levels::Vector) where {T}
+    index = map!(t -> ismissing(t) ?
+                    findfirst(ismissing, levels) :
```
nalimilan: Really weird indentation here.

nalimilan:
> It looks like there's an issue somewhere here with levels/sorting. I'm getting a mismatch between group labels and the actual data plotted using Geom.subplot_grid.

I don't really understand what's going on in the figure you showed, but if that's related to the order of levels then it might be due to the fact that `IndirectArray` does not support a custom ordering of levels? I've just realized that it could be a significant problem: with this PR, is the custom ordering of levels preserved when plotting for `CategoricalArray` variables? That's one of the main advantages of this type, and `PooledDataArray` supported it.

src/scale.jl
andreasnoack: It is the first four space aligned position after the `(`. Arguments are aligned to the `(` but this splits a ternary instead of two arguments so it should be indented. The call is not super easy to read as is but the indentation makes it easier.

andreasnoack:
I'll have to take a closer look. Do you have a minimal example to reproduce this? My guess is that I'd have to be more careful when converting a CategoricalVector to an indirect Vector such that the `values` end up in the right order.

nalimilan:
Sorry, I don't use Gadfly, but I guess something like [this](http://gadflyjl.org/stable/tutorial.html#Discrete-scales-1) should do if you set a custom non-alphabetical order for countries (e.g. grouped by regions). Note that it may already be working, I just happened to think about it when reading the description of the issue above.

src/scale.jl
nalimilan: AFAICT there are only three spaces, not four (assuming the count starts *after* `(`, i.e. starts with the `t`). More importantly, what bugs me is that the second operand isn't aligned with the first one.

kleinschmidt:
@nalimilan the problem with the figure I showed is that the labels on the facets (\alpha=0.01, 0.1, 1, 10) are wrong.  with the PR they've been circshifted by one, while the facets are still plotted in the same order (just the labels have changed).  The values that are used to group the facets are a data frame column that's an `Array{String}`.

I will look into this more later and figure out a minimal example but don't have time right at the moment.

bjarthur:
i merged a PR which requires some conflicts to be resolved with this one.  sorry, but they are small.  good thing is that regression testing is even easier now!

andreasnoack:
@kleinschmidt I've pushed a possible fix for this. Would be great if you could try it out.

@bjarthur No problem. I've rebased.

bjarthur:
before we merge this, would it make sense to alter REQUIRE to put a ceiling on DataFrames  of 0.11.4?  if so, i'd say let's do that ASAP.

also, why were breaking changes in DataFrames made on a point release?  shouldn't they have bumped the minor version?

andreasnoack:
> would it make sense to alter REQUIRE to put a ceiling on DataFrames of 0.11.4?

I'm not sure I follow. Why would you put a ceiling there? We could try to put a ceiling at the next breaking changing version but DataFrames doesn't follow SEMVER (yet) so it for to know if it 0.14, 0.17, or 1.0.

The breaking change in `DataFrames` happened at 0.11.0 but there have been some bug fixes after that. I think I decided to use 0.11.4 because it has a fix for the deprecated `readtable` function.

I'll try to finish the IndirectArrays PR soon such that we can have this merged.

src/geom/line.jl
```diff
@@ -131,7 +131,7 @@ function render(geom::LineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics)
                       svgclass(classes))
 
     elseif length(aes.color) == 1 &&
-            !(isa(aes.color, IndirectArray) && length(levels(aes.color)) > 1)
+            !(isa(aes.color, IndirectArray) && length(filter(!ismissing, aes.color.values)) > 1)
```
nalimilan: Could also use `count(!ismissing, aes.color.values)` to avoid an allocation. Same below.

kleinschmidt:
LGTM now @andreasnoack, thanks for your work on this!

src/scale.jl
```diff
@@ -251,45 +253,51 @@ function apply_scale_typed!(ds, field, scale::ContinuousScale)
...
-        return values
-    else
-        return PooledDataArray(values, convert(Vector{eltype(values)}, levels))
-    end
+discretize_make_pda(values::IndirectArray)         = values
+discretize_make_pda(values::IndirectArray, ::Void) = values
+discretize_make_pda(values::IndirectArray, levels) =
+    IndirectArray(values, convert(Vector{eltype(values)}, levels))
+
+discretize_make_pda(values::CategoricalArray)         = discretize_make_pda(values, unique(values))
```
nalimilan: The ordering returned by `levels(values)` should be preserved here, that's the whole point of categorical arrays. `intersect(vcat(levels(values), missing), unique(values))` should work and will sort missing values at the end.

src/scale.jl
```diff
@@ -251,45 +253,51 @@ function apply_scale_typed!(ds, field, scale::ContinuousScale)
...
+    IndirectArray(map(t -> findfirst(levels, t), values), levels)
 
-function discretize_make_pda(values::DataArray, levels=nothing)
-    if levels == nothing
-        return PooledDataArray(values)
-    else
-        return PooledDataArray(convert(DataArray{eltype(levels)}, values), levels)
-    end
-end
+discretize_make_pda(values::DataArray)         = IndirectArray(values)
```
nalimilan: I don't think this specialization for `DataArray` is needed, is it? Even when missing values need special treatment, `AbstractArray{>:Missing}` is usually enough.

src/scale.jl
```diff
@@ -488,7 +496,8 @@ function apply_scale(scale::DiscreteColorScale,
     for (aes, data) in zip(aess, datas)
         data.color === nothing && continue
         for d in data.color
-            isna(d) || push!(levelset, d)
+            # ismissing(d) ||
```
nalimilan: ?

src/statistics.jl
```diff
@@ -983,8 +983,8 @@ function apply_statistic(stat::BoxplotStatistic,
         end
 
         if aes.color !== nothing
-            aes.color = PooledDataArray([c for (x, c) in groups],
-                                        levels(aes.color))
+            aes.color = IndirectArray([c for (x, c) in groups],
+                                        filter(!ismissing, aes.color.values))
```
nalimilan: Wrong indentation.

src/scale.jl
```diff
@@ -251,45 +253,51 @@ function apply_scale_typed!(ds, field, scale::ContinuousScale)
...
+                        findfirst(s -> !(ismissing(s) || s != get(t)), levels),
+                 Array{UInt8}(size(values)...), values)
+    return IndirectArray(index, convert(Vector{T},levels))
 end
+function discretize_make_pda(values::CategoricalArray{T}, levels::CategoricalVector{T}) where T
+    _levels = map!(t -> ismissing(t) ? t : get(t), Vector{T}(length(levels)), levels)
+    discretize_make_pda(values, _levels)
+end
+
+# These methods convert WeakRefStringArrays to Vector{String} and shouldn't really be necessary
```
nalimilan: Should be OK with recent releases of WeakRefStrings, right?

nalimilan:
@kleinschmidt Have you figured out the reason for the bug you mentioned above?

REQUIRE
```diff
@@ -1,10 +1,10 @@
-julia 0.5
+julia 0.6
```
bjarthur: should stop testing on 0.5 if Gadfly is dropping support for it, particularly since METADATA no longer supports it.

bjarthur:
big picture question here.  it seems that PooledDataArrays were originally used to maintain an ordering of levels, which seems to still be needed.  given that, was it ever considered to used https://github.com/JuliaComputing/PooledArrays.jl instead of IndirectArrays?  it doesn't wrap values like CategoricalArrays does, and it maintains a sorted ordering of levels.  so no need to modify IndirectArrays to support setindex, nor impose our own ordering on it's pool.  seems easier.  am i missing something?

Mattriks:
Since we are on the topic of other options, is it possible to stick with `PooledDataArray`, by copying, updating and maintaining the `PooledDataArray` code from `DataArrays.jl` within `Gadfly`?  Note that `PooledDataArray` needs only to be available internally in `Gadfly`, it doesn’t need to be available externally.    The advantage of this is to future proof Gadfly from changes in external packages, which is the cause of this current PR.

andreasnoack:
> was it ever considered to used https://github.com/JuliaComputing/PooledArrays.jl instead of IndirectArrays?

I did consider that first but, as @nalimilan mentions, the build fundamentally on the values array being sorted which wouldn't work here.

> is it possible to stick with `PooledDataArray`

I new version of IndirectArrays has now been merged so I just need to update this PR with the feedback from the reviews and we should be able to update Gadfly.

bjarthur:
@nalimilan thanks so much for your help with this issue.  great to have https://github.com/JuliaArrays/IndirectArrays.jl/pull/12 merged.  do you by chance have any idea when julia 0.6.3 will be released?  would be great to have https://github.com/JuliaLang/julia/pull/25832 backported so we no longer have to use the `compilecache=no` flag with this PR.

nalimilan:
Sorry, I have no idea. Maybe we could work around the ambiguity issues by defining specific methods temporarily?

src/scale.jl
andreasnoack: I'd prefer not to change this now. We can do another round of cleanup in a new PR.

src/scale.jl
andreasnoack: Probably not, but it could happen so I think we should keep them.

nalimilan: `WeakRefStringArray`s now return `String` objects when indexed for several releases, that's why I said that's probably no longer needed (as the comment says).

andreasnoack:
@nalimilan For some reason I can't reply inline to

> Better do something like:
```julia
mapping = coalesce.(indexin(CategoricalArrays.index(values.pool), levels), 0)
pushfirst!(mapping, coalesce(findfirst(ismissing, levels), 0))
mapping .+= 1
index = [mapping[x+1] for x in values.refs]
```
I tried this but it is not functionally equivalent so tests failed. I'll suggest that we merge the current version and improve performance later.

src/scale.jl
andreasnoack: This code path is not preserving order. There is a separate branch in `discretize` taken when the ordering should be preserved. We might want to change this later but I don't think such a reorganization belongs here.

src/scale.jl
andreasnoack: It might be possible to define these for more abstract types but I'm not sure it will always do the right thing so, for now, I prefer strict signatures.

src/scale.jl
andreasnoack: This should just have been deleted.

src/scale.jl
andreasnoack: `WeakRefStringArrays` are still passed to Gadfly and no other signature currently matches.

src/scale.jl
nalimilan: The existing `PooledDataArray` method did preserve order since it passed the `levels` argument to the PDA constructor. So I think you should keep the current `level=nothing` pattern, using the code I posted above to fill `levels` when it's `nothing`.

src/scale.jl
nalimilan: How about changing the `::Vector` method to `::AbstractVector`, and using the `IndirectArray` convenience constructor you've just added? That would allow getting rid of the dependencies on WeakRefStrings and DataArrays, which really shouldn't be here since we only use the `AbstractArray` interface.

src/scale.jl
```diff
@@ -251,45 +253,52 @@ function apply_scale_typed!(ds, field, scale::ContinuousScale)
...
+discretize_make_ia(values::CategoricalArray, ::Void) = discretize_make_ia(values)
+function discretize_make_ia(values::CategoricalArray{T}, levels::Vector) where {T}
+    index = map!(t -> ismissing(t) ?
+                    findfirst(ismissing, levels) :
+                    findfirst(s -> !(ismissing(s) || s != get(t)), levels),
+                 Array{UInt8}(size(values)...),
+                 values)
+    return IndirectArray(index, convert(Vector{T},levels))
 end
+function discretize_make_ia(values::CategoricalArray{T}, levels::CategoricalVector{T}) where T
```
nalimilan: Really there should be no need for this special method. `::AbstractArray` should work equally for this one and the one above.

src/scale.jl
```diff
@@ -251,45 +253,52 @@ function apply_scale_typed!(ds, field, scale::ContinuousScale)
...
-function discretize_make_pda(values::Vector, levels=nothing)
-    if levels == nothing
-        return PooledDataArray(values)
-    else
-        return PooledDataArray(convert(Vector{eltype(levels)}, values), levels)
-    end
-end
+discretize_make_ia(values::Vector)         = IndirectArray(values)
+discretize_make_ia(values::Vector, ::Void) = IndirectArray(values)
+discretize_make_ia(values::Vector, levels) =
```
nalimilan: `IndirectArray`s behave differently from PDAs if some levels are omitted. Do you think this is used anywhere (i.e. to skip some levels)?
```julia
julia> PooledDataArray(["a", "b", "c"], ["a", "c"])
3-element DataArrays.PooledDataArray{String,UInt32,1}:
 "a"    
 missing
 "c"    

julia> IndirectArray(["a", "b", "c"], ["a", "c"])
ERROR: BoundsError: attempt to access 2-element Array{String,1} at index [UInt8[0x01, 0x00, 0x02]]
```

nalimilan:
> I tried this but it is not functionally equivalent so tests failed. I'll suggest that we merge the current version and improve performance later.

Hmm, looks like there was a bug in the code I posted. Here's a version which seems to give correct results (at least compared with the current implementation in the PR, which throws an error for missing levels)
```julia
    mapping = coalesce.(indexin(CategoricalArrays.index(values.pool), levels), 0)
    unshift!(mapping, coalesce(findfirst(ismissing, levels), 0))
    index = [mapping[x+1] for x in values.refs]
    any(iszero, index) && throw(ArgumentError("values not in levels encountered"))
    return IndirectArray(index, convert(Vector{T},levels))
```

This is not just a small optimization: calling `findfirst` for each entry is going to be incredibly slow and a clear regression from PDAs.

andreasnoack:
> This is not just a small optimization: calling `findfirst` for each entry is going to be incredibly slow and a clear regression from PDAs.

I'm not convinced that is true in real usecases. The complexity os worse but `n` is small.
```julia
julia> f(x, p) = map(t -> findfirst(p, t), x);

julia> ps = 2.^(1:8)
8-element Array{Int64,1}:
   2
   4
   8
  16
  32
  64
 128
 256

julia> map(ps) do pi
         p = shuffle(1:pi); y = rand(p, 10000); gc()
         t1 = @elapsed i1 = f(y, p)
         t2 = @elapsed i2 = indexin(y, p)
         @assert i1 == i2
         return (pi, t1, t2)
       end
8-element Array{Tuple{Int64,Float64,Float64},1}:
 (2, 0.000133984, 0.000251059)
 (4, 0.000209158, 0.000260757)
 (8, 0.000198923, 0.000266283)
 (16, 0.000399805, 0.000318315)
 (32, 0.000436547, 0.000361112)
 (64, 0.000684135, 0.000497858)
 (128, 0.000827828, 0.000349345)
 (256, 0.00152649, 0.000314503)
```
So the crossover is between 16 and 32 which I think is more levels than you usually have.

I'm willing to try your new version but if the tests fail I won't spend more time on this method. You can play around with cleaning this up afterward if you believe it is necessary.

src/scale.jl
andreasnoack: I'm not completely sure but it looks like Gadfly typically constructs the union of available levels. If Gadfly made use of this behavior, I'd also expect that at least one of the tests would have caught it.

src/scale.jl
andreasnoack: Most of the time spent (which now was quite a bit) was spent on making sure that no wrapped types made it through to `Compose`. The problem with `AbstractArray` is that you don't know if the elements are wrapped on not. Concrete methods handle the cases explicitly. We might be able to clean this up but these are internal methods so that can be done later if anybody feels for doing it.

src/scale.jl
andreasnoack: Again, we don't want any wrapped types to escape to `Compose` which can be difficult when using `AbstractArray` signatures. Also, note that there isn't a dependency on `WeakRefStrings `. I'm fine with working towards dropping the `DataArrays` dependency but not in this PR.

src/scale.jl
nalimilan: Yeah, I guess we'll soon find out if that's not the case anyway. :-)

nalimilan:
> So the crossover is between 16 and 32 which I think is more levels than you usually have.

This benchmark misses the point. The code I suggested doesn't call `indexin` on the values vector, only on the levels (i.e. on a vector of length typically below 50). It then uses direct indexing into a vector of codes, which is very fast (and can use SIMD). That pattern is what makes categorical/pooled/indirect arrays interesting for performance in many situations.

andreasnoack:
This version worked so the PR has been updated.

Another thing. Gadfly checks out `Compose` in `.travis.yml` and this version actually doesn't work with latest release of `Compose`. I have a vague memory of this being discussed but I don't recall when, where or the details. Anyway, I think we should tag `Compose` and remove the line that checks out `Compose`.

bjarthur:
wait, what's wrong with the latest version of Compose?  i just pushed a bunch of stuff to master last night in preparation for merging https://github.com/GiovineItalia/Gadfly.jl/pull/1037

andreasnoack:
> wait, what's wrong with the latest version of Compose?

[This method here](https://github.com/GiovineItalia/Compose.jl/blob/e5e4f54419694df6c41b9c0a2e6fa351cd4985ca/src/pgf_backend.jl#L122) doesn't match the [current invokation](https://github.com/GiovineItalia/Gadfly.jl/blob/a5b01ff9eb472fd127ea5f352bf99daf111f229f/test/runtests.jl#L59) in Gadfly's test.

andreasnoack:
> It would also be nice if @kleinschmidt could confirm whether he still sees the bug he mentioned above.

I read https://github.com/GiovineItalia/Gadfly.jl/pull/1090#issuecomment-364930613 as a confirmation.

bjarthur:
i don't believe you have the current master of Compose checked out, as [this](https://github.com/GiovineItalia/Compose.jl/blob/master/src/pgf_backend.jl#L122) is what those methods look like.

andreasnoack:
> i don't believe you have the current master of Compose checked out

I do have that but my point is that I shouldn't. I https://github.com/GiovineItalia/Gadfly.jl/pull/1090#issuecomment-373067584 I was talking about the latest *released* version of Compose.

tlnagy:
> I do have that but my point is that I shouldn't.

Why is that? In the past we've always moved `Compose` and `Gadfly` in lockstep so it makes sense to validate whether or not Gadfly works on top of the latest master of Compose.

andreasnoack:
> Why is that? In the past we've always moved Compose and Gadfly in lockstep so it makes sense to validate whether or not Gadfly works on top of the latest master of Compose.

Then why have two packages? If the code is so tightly connected then maybe it should be two separate packages. If they are not so tightly connected then normal versioning should be sufficient although it might be slightly cumbersome at times. I can't think other Julia packages that do anything like this.

bjarthur:
yea, it's weird, i [agree](https://github.com/GiovineItalia/Gadfly.jl/issues/1007), but consider that there are [28 other packages](https://pkg.julialang.org/?pkg=Compose#Compose) that require Compose.  so we can't unmerge them (they were unified originally i believe).  yet the coupling is so tight that we have to keep them in lockstep.

andreasnoack:
I don't think it is impossible to merge packages. We plan on doing that with `StatsFuns` and `Distributions`. Some package would have to get upper bounds but that happens frequently.

Anyway, let's take that discussion another day. I guess this PR is ready.

test/runtests.jl
```diff
@@ -108,3 +108,9 @@ close(output)
 if prev_theme !== nothing
     ENV["GADFLY_THEME"] = prev_theme
 end
+
+if !haskey(ENV, "TRAVIS") && !isinteractive() &&
+            !isempty(readdir(joinpath((@__DIR__),"cachedoutput"))) &&
+            !isempty(readdir(joinpath((@__DIR__),"gennedoutput")))
+    run(`$(Base.julia_cmd()) $(joinpath(@__DIR__, "compare_examples.jl"))`)  # `include`ing causes it to hang.
+end
```
bjarthur: this looks like an oversight in a rebase.  we don't want this here because it causes problems with PkgEval, which does *not* have the TRAVIS environment variable set.  when testing locally, developers will simply have to run the regression tests manually.

.travis.yml
```diff
@@ -6,11 +6,17 @@ julia:
...
+  - julia: nightly
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia -e 'Pkg.clone(pwd()); Pkg.build("Gadfly");'
   - julia -e 'Pkg.checkout("Compose")'
-  - julia --check-bounds=yes -e 'Pkg.test("Gadfly", coverage=true)'
+  # because of a bug in Julia 0.6.2 we temporarily use --compilecache=no.
+  # The bug is fixed on master and will be included in 0.6.3 so when it
+  # has been releases, we can bump the requirement and remove the flag.
+  - julia --check-bounds=yes --compilecache=no -e 'Pkg.test("Gadfly", coverage=true)'
```
bjarthur: it was [suggested](https://github.com/GiovineItalia/Gadfly.jl/pull/1090#issuecomment-372654098) that some temporary methods could be added which would alleviate the need to use `compilecache`.  do you think that's possible?  if so, that would be a much better solution.

test/runtests.jl
andreasnoack: Indeed. I've cleaned this up now.

.travis.yml
andreasnoack: This happens to methods in `IndirectArrays` as well as `WeakRefStrings` so we'd have to add temporary methods to at least two packages even though the methods shouldn't be needed and the bug can only be triggered by Gadfly. Releasing a version that will cause ambiguity errors is also really bad though. We only have bad options.

andreasnoack: Let me try to add the methods to the two packages. If tests pass with the methods defined, it might be worth doing even though I'd much prefer a 0.6.3 release to doing this.

andreasnoack: Meanwhile, I think it would make sense to merge this as is but holding off on a new release until we have looked more into this issue.

nalimilan: What are the problematic methods? Possibly related: https://github.com/JuliaData/WeakRefStrings.jl/pull/26

andreasnoack: Yes. That might be one of them. I'll try to check out master of WeakRefStrings and see how it goes.

andreasnoack: With WeakRefString master and https://github.com/JuliaArrays/IndirectArrays.jl/pull/13 all Gadfly tests pass with precompilation enabled. So it looks like we can actually work around this in a reasonable way.

andreasnoack:
I've now bumped the requirements and remove `--compilecache=no` from `.travis.yml`.

.travis.yml
```diff
@@ -6,10 +6,16 @@ julia:
...
+matrix:
+  allow_failures:
+  - julia: nightly
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia -e 'Pkg.clone(pwd()); Pkg.build("Gadfly");'
   - julia -e 'Pkg.checkout("Compose")'
+  # because of a bug in Julia 0.6.2 we temporarily use --compilecache=no.
+  # The bug is fixed on master and will be included in 0.6.3 so when it
+  # has been releases, we can bump the requirement and remove the flag.
```
bjarthur: this comment no longer needs to be here.  otherwise looks good.  all the regression tests pass?

bjarthur:
i've found quite a few regression errors with `julia test/compare_examples.jl --bw js.svg`.  specifically, the following file all had important differences between master and this PR:  line_histogram, issue82, dodged_discrete_histogram, stacked_continuous_histogram, and subplot_grid.

bjarthur:
two more things:  should remove julia 0.5 PkgEval badge from README, and should make the docs locally to ensure they all work.  at least one is broken with the new DataFrames.

nalimilan:
> In the histograms, missing is now showing up as a level. This happens because missing is now a value in the pool instead of sentinel index. I'll figure out how to get the old behavior back. (Actually, I could imagine it would be useful to be able to get the missing values as a level in the plot but that it for another issue/PR.)

Indeed, the old code called `levels` on PDAs, which skips missing values. You should be able to continue using `levels`, which is now defined in Missings.jl and also skips missing values. This function is documented to return values in a meaningful order (increasing/lexicographic by default) rather than in the order of appearance, and any custom array type can override it to follow a custom ordering (e.g. `CategoricalArray`). For `CategoricalArray`, it looks like `intersect(push!(levels(values), missing), unique(values))`  could just be `levels(values)`, just like it was for PDAs.

cecileane:
Any update on this? The upper bound on Dataframe < 0.11.0 causes issues for [PhyloNetworks](https://github.com/crsl4/PhyloNetworks.jl). We are eagerly awaiting resolution of this issue.

tlnagy:
Gadfly's upper bound here breaks most of the `DataStreams.jl` ecosystem, which is really unfortunate. It would be great to have this merged ASAP.

andreasnoack:
Today, I've had some time to look at this. The main issue was Gadfly's handling of missing values. I had missed https://github.com/GiovineItalia/Gadfly.jl/blob/a111e3835fa1a9413fb47b9bed9288c8f7e949c5/src/scale.jl#L511-L518 which in the middle of applying the scales, also removes missing values from color scales only by exploiting implementational details of how missing values encoded in `PooledDataArray`s. I've tried to make it clearer but it would be great to find a more robust and maintainable way of handling missing values in Gadfly.

With the new commits, the issue with missings showing up in the plots should be fixed. I get 34 plots where the `tex` files have changed but I've visually inspected all of the corresponding `svg`s and I'm unable to distinguish the old from the new ones.

The only genuinely different plot is `issue82` and here I'd like to start a discussion. The difference is that a `PooledDataArray{<:Number}` used to be classified as `:numerical` data but I've currently made it such that data in `CategoricalArray` is classified as, surprise,  `:categorial` regardless of the value type. I think this is the better behavior and ensures that no `CategoricalValue`s float around. With this version, it is not possible to apply a continuous scale to an aesthetic from a `CategoricalArray` which I think makes sense and I've added an informative error message in https://github.com/GiovineItalia/Gadfly.jl/pull/1090/commits/905db904b6b0cbf8932ebfb8f38dce3a95cf581f. Are people okay with this change?

kleinschmidt:
That seems very reasonable to me.  It would be confusing to me as a user if I had a numeric categorical array that was plotted as a continuous value; I'd like to have a way to plot numeric things as categorical (e.g., integer indices) and converting it to a categorical array is a sensible way to support that.

andreasnoack:
The previous test error was because of an improvement in `RData`. Hopefully, tests will now pass.

REQUIRE
```diff
@@ -16,3 +16,10 @@ Loess
 Showoff 0.0.3
 StatsBase
 Juno
+IndirectArrays 0.4.1
+Missings
+# Gadfly doesn't use WeakRefString directly but because of a but in Julia 0.6.2, Gadfly
```
tlnagy: bug not but

tlnagy:
Old behavior:

```julia
a = DataFrame(diff = PooledDataArray(Float64[1,2,3,3,3,4,3,2]))
plot(a, x="diff", Geom.histogram)
```



New behavior:

```julia
a = DataFrame(diff = CategoricalArray(Float64[1,2,3,3,3,4,3,2]))
plot(a, x="diff", Geom.histogram)
```




I think the new behavior makes sense with the transition from `PooledDataArrays` to `CategoricalArrays`. The latter is what I would expect for categorical data. And you can get the old behavior back if you just use an `Array`

```julia
a = DataFrame(diff = Float64[1,2,3,3,3,4,3,2])
plot(a, x="diff", Geom.histogram)
```

tlnagy:
I'm okay with merging since tests are passing and there aren't any regressions. @bjarthur, thoughts?

src/scale.jl
```diff
@@ -173,6 +174,9 @@ function apply_scale(scale::ContinuousScale,
     for (aes, data) in zip(aess, datas)
         for var in scale.vars
             vals = getfield(data, var)
+            if vals isa CategoricalArray
```
nalimilan: Is there a strong reason to throw an error here? I fully agree that by default `CategoricalArray` should be treated as categorical even if it contains numeric values (that was the rationale for moving from `PooledDataArray`), but if the user explicitly requests a continuous scale, why not accept it? Or is it because the code fails later?

andreasnoack: > is there a strong reason to throw an error here?

Yes. It won't work. The `CategorialcalValue`s will cause failures sooner or later so I thought an informative error would be more helpful. The specific error in the example I tried is a missing `isfinite(::CategoricalValue)`. As mentioned previously, the core of this PR is to avoid that `CategoricalValue`s escape to `Compose`. In the discrete case, all data vectors go through `discretize` where `CategoricalArray`s are converted to `IndirectArrays` to avoid `CategoricalValue`s but in the continuous case there isn't a similar function we can use for making sure all `CategoricalArray`s are converted.

> that was the rationale for moving from `PooledDataArray`

I don't really understand why it wouldn't be exactly the same thing for `PooledDataArray`s. I'd also expect them to be treated as categorical.

nalimilan: OK. That can always be improved later anyway if somebody needs it. Thanks for finishing this porting work!

> I don't really understand why it wouldn't be exactly the same thing for PooledDataArrays. I'd also expect them to be treated as categorical.

Yeah, but there had always been an ambiguity, as PDAs were also used as an efficient storage type (e.g. they supported math operations). Now the distinction should be clearer, if only because of the `CategoricalArray` name.

bjarthur:
`subplot_grid` and `subplot_grid_free_axis` still show regression differences for me.

andreasnoack:
> `subplot_grid` and `subplot_grid_free_axis` still show regression differences for me.

I get the plots below. Indeed the years have flipped but I think the new version is the correct one.

```julia
julia> barley = dataset("lattice", "barley");

julia> levels!(barley[:Year], ["1931", "1932"]);

julia> barley[(barley[:Year] .== "1931") .& (barley[:Site] .== "Morris") .& (barley[:Variety] .== "No. 475"),:]
1×4 DataFrames.DataFrame
│ Row │ Yield │ Variety │ Year │ Site   │
├─────┼───────┼─────────┼──────┼────────┤
│ 1   │ 22.6  │ No. 475 │ 1931 │ Morris │

julia> barley[(barley[:Year] .== "1932") .& (barley[:Site] .== "Morris") .& (barley[:Variety] .== "No. 475"),:]
1×4 DataFrames.DataFrame
│ Row │ Yield   │ Variety │ Year │ Site   │
├─────┼─────────┼─────────┼──────┼────────┤
│ 1   │ 44.2333 │ No. 475 │ 1932 │ Morris │
```

### cached
subplot_grid


### genned
subplot_grid

bjarthur:
agreed!  i've tagged gadfly and compose.  as soon as that's merged into METADATA i'll merge this.

bjarthur:
@kleinschmidt @cecileane have you two had a chance to checkout Gadfly master since this PR merged?  just want to make sure all is okay before we tag a release.  thanks!

kleinschmidt:
I’ll have a look later today.

cecileane:
Thanks @bjarthur! Yes I checked out Gadfly, and I first ran into a bunch of problems (see this [issue](https://github.com/cecileane/PhyloPlots.jl/issues/1)). I didn't figure out why, but things are now working (on a different computer) after checking out the master branch of Compose 👍 . After you guys publish a new version of Gadfly (and Compose?), I will add Gadfly back into [PhyloPlots](https://github.com/cecileane/PhyloPlots.jl).

cecileane:
well, my earlier comment was using julia v0.6.1. Things don't work with julia v0.6.2. At compilation, I get the same error as mentioned [here](https://github.com/cecileane/PhyloPlots.jl/issues/1), with a very cryptic stacktrace:

```
ERROR: LoadError: MethodError: no method matching convert(::Type{AssertionError}, ::String)
Closest candidates are:
  convert(::Type{Any}, ::ANY) at essentials.jl:28
  convert(::Type{T}, ::T) where T at essentials.jl:29
```
and final error:
```
MethodError(Core.Inference.convert, (AssertionError, "invalid age range update"), 0x0000000000000ac6)
```
This is using the master version of Gadfly and Compose, and adding back "using Gadfly" in [PhyloPlots.jl](https://github.com/cecileane/PhyloPlots.jl/blob/master/src/PhyloPlots.jl). Not even including the file that uses Gadfly functions. Any idea welcome!

kleinschmidt:
All is well for my use case! Thanks for working so hard on this everyone!

cecileane:
yes indeed, thanks @andreasnoack. It looks like the same kind of error that @mauro3 showed, and it does seem sensitive to using lots of modules together.

tlnagy:
Is `DataArrays` still needed in REQUIRE? https://github.com/GiovineItalia/Gadfly.jl/blob/e36d55c5bf42921b0013ed8a63fa749415b84d86/REQUIRE#L8

bjarthur:
so just to clarify, with this now merged into master and IndirectArrays modified, does one need to use the `--compilecache=no` flag when using Gadfly?

andreasnoack:
> does one need to use the --compilecache=no flag when using Gadfly?

No. Things should work and the flag is not present in `.travis.yml`. The only regression that I'm aware of is https://github.com/GiovineItalia/Gadfly.jl/issues/1130. I think I know how to fix it but I'm off for a week of vacation on Sunday so might not get around fixing it before that. It is mainly a matter of changing the part of https://github.com/GiovineItalia/Gadfly.jl/blob/12327dd63fb5d84636d147b9d2d82eb76f8b6b1c/src/misc.jl#L408-L414 that throws and instead convert the zeros in `index` to `n+1` and add `missing` to `levels`. However, https://github.com/GiovineItalia/Gadfly.jl/blob/12327dd63fb5d84636d147b9d2d82eb76f8b6b1c/src/misc.jl#L397-L398 should probably also be changed similarly in case the input is e.g. a `Vector{String}`.

PR #1095, opened 2018-02-01
add example with a discrete scale and perserved order

### LINKED ISSUE #1079: fix order for Geom.violin
I'm trying to create a violin plot from a DataFrame, with x=Col.index and y=Col.value. It works fine, except that the order of the columns is not preserved in the plot. Is there a way to fix the order? Thanks in advance!

tlnagy:
fixes #1079

### DIFF
--- a/docs/src/lib/scales/scale_x_discrete.md
+++ b/docs/src/lib/scales/scale_x_discrete.md
@@ -32,6 +32,7 @@ easiest way to get that data to plot appropriately.
 
 ```@setup 1
 using RDatasets
+using DataFrames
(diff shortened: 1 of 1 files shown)


tlnagy:
@bjarthur Do you think tests are failing on 0.6 due to DataFrames?

bjarthur:
seems likely given the `CategoricalArray` error that beeswarm is throwing.  how about we add `--compilecache=no` to .travis.yml, re-run the CI for https://github.com/GiovineItalia/Gadfly.jl/pull/1090 and merge if it passes?

PR #1097, opened 2018-02-12
upperbound DataFrames to isolate breaking change

bjarthur:
this should work, no?, to fix Gadly on julia 0.6?  if so, i'll merge and then tag a version.  let me know b/c i'm not completely sure.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,7 +4,7 @@ Compat 0.18.0
 Compose 0.5.2
 Contour 0.1.1
 CoupledFields
-DataFrames 0.4.2
+DataFrames 0.4.2 0.11.0
 DataArrays
 DataStructures
 Distributions


bjarthur:
tests pass.  i'll merge and tag tomorrow unless someone advises against it.

tlnagy:
I think this should, since all the breaking changes were pushed with DataFrames v0.11.0. The one problem with this would be that Gadfly would cause a conflict with DataFrames installations v0.11.0+ instead of throwing a bunch of deprecation warnings.

bjarthur:
hmm.  i had thought that 0.11.0+ would throw errors too.  to try to reproduce, i freshly installed Gadfly with `JULIA_PKGDIR=.julia-test julia` and then `Pkg.add("Gadfly")`.  to my surprise, DataFrames 10.1 was installed.  why not 11.5?

Gadfly's tests succeeded with 10.5, as expected.  to confirm that 11.5 would break them, i tried to pin DataFrames, but got `ERROR: Gadfly can't be installed because it has no versions that support 0.6.2 of julia.`.  can someone please explain?  i'm totally confused.

bjarthur:
beeswarm, boxplot, and colored_points all failed with Gadfly master and DataFrames 0.11.5.  all tests succeed when pinning the latter to 0.10.1 though.  so i'll merge and tag.

PR #1099, opened 2018-02-16
make Geom.point the default for layers too

### LINKED ISSUE #1062: Combining plots with differing number of points
I wanted to do something along the lines of 
`plot(layer(x=X[1:10],y=Y[1:10]), layer(x=X[IndexSet],y=Y[IndexSet]))`
where `length(IndexSet)` is less than 10. Gadfly dies with 
`MethodError: no method matching render(::Gadfly.Geom.Nil, ::Gadfly.Theme, ::Gadfly.Aesthetics)`
I experimented around and I find the same error anytime the layers have a differing numbers of points.

Is that intended behavior? I didn't see any mention of constraints of this nature in the documentation.

tlnagy:
we already have Geom.point as the default when there are no layers. this
extends this to if no geometry is specified for a layer. Closes #1062

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -424,6 +424,7 @@ function render_prepare(plot::Plot)
 
             evalmapping!(layer.mapping, layer.data_source, datas[i])
         end
+        if isa(layer.geom, Geom.Nil); layer.geom = Geom.point(); end # see #1062
     end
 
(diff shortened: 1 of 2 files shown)


tlnagy:
@bjarthur Any thoughts on this? This seems reasonable to me.

bjarthur:
yes, ofc!  how about a test?

Issue #1069, opened 2017-12-21
Geom.hline not supported in Geom.subplot_grid?


ferris-wufei:
I was trying to add Geom.hline on top of Geom.point within Geom.subplot_grid, but failed with the following code and error msg. Did I do anything wrong? Or it's not supported?

```julia
using DataFrames
using Gadfly
using RDatasets

# Gadfly                        0.6.4
# RDatasets                     0.2.0
# DataFrames                    0.9.1

# 1st try, place Geom.hline within Geom.subplot_grid
plot(dataset("datasets", "iris"), x="SepalLength", y="SepalWidth", xgroup="Species", yintercept=[2.5, 4.0], Geom.subplot_grid(Geom.point, Geom.hline(style=:dot)))

The following aesthetics are required by Geom.hline but are not defined: yintercept


 in assert_aesthetics_defined(::String, ::Gadfly.Aesthetics, ::Symbol, ::Vararg{Symbol,N}) at /Users/wufei/.julia/v0.5/Gadfly/src/aesthetics.jl:156
 in render(::Gadfly.Geom.HLineGeometry, ::Gadfly.Theme, ::Gadfly.Aesthetics) at /Users/wufei/.julia/v0.5/Gadfly/src/geom/hvabline.jl:32
 in render(::Gadfly.Geom.HLineGeometry, ::Gadfly.Theme, ::Gadfly.Aesthetics, ::Array{Gadfly.Aesthetics,1}, ::Array{Gadfly.Data,1}, ::Dict{Symbol,Gadfly.ScaleElement}) at /Users/wufei/.julia/v0.5/Gadfly/src/geometry.jl:42
 in (::Gadfly.##117#119{Dict{Symbol,Gadfly.ScaleElement}})(::Tuple{Gadfly.Layer,Gadfly.Aesthetics,Array{Gadfly.Aesthetics,1},Array{Gadfly.Data,1},Gadfly.Theme}) at ./<missing>:0
[...]

PR #1101, opened 2018-02-19
Examples for docs

### LINKED ISSUE #1091: Feature request: control over ordering of factors in stacked bar chart
I have created a stacked bar chart and would like to reverse the order in which the factors appear in each bar. Is there an easy way to do this? If not, it would be nice to have an option like `Geom.bar(position=:stack, rev=true)`.

Mattriks:
Adds plot examples from #1091 and #1069 to docs (closes #1091).


bjarthur:
thx!

PR #1102, opened 2018-02-20
don't automatically call compare_examples; add flags

bjarthur:
dang.  regression tests are still broken.

who would've thought that `runtests.jl` was non-interactive when called by `Pkg.test` and interactive when `include`d:

```
$ head -1 test/runtests.jl 
info(isinteractive());  exit()
```

```
julia> Pkg.test("Gadfly")
INFO: Computing test dependencies for Gadfly...
INFO: No packages to install, update or remove
INFO: Testing Gadfly
INFO: false
INFO: Gadfly tests passed
INFO: No packages to install, update or remove

julia> include("test/runtests.jl")
INFO: true
```

this PR should really fix `compare_examples.jl` now.  it also adds some nice flags to control its behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@ test/gennedoutput.html
 test/cachedoutput.html
 test/gennedoutput
 test/cachedoutput
+test/diffedoutput
 doc/html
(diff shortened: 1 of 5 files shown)


tlnagy:
Do you think this will fix package eval too?

Kinda annoying that this is front and center:

image

bjarthur:
the last pkgeval [log](https://pkg.julialang.org/logs/Gadfly_0.6.log) says the problem is with `test/testscripts/subplot_grid_free_y_1.jl`.  works fine locally, so not sure what's going on.

there seems to be some non-determinism involved as, for example, Gadfly v0.6.1 [tests](https://pkg.julialang.org/detail/Gadfly.html) failed on 2017-04-25, passed on 2017-05-03, and then started failing again on 2017-05-04.  perhaps this is due to a dependency.

a better error message sure would help.  better yet, access to one of the test boxes.

PR #1103, opened 2018-02-25
Geom.ellipse

Mattriks:
`Geom.ellipse` calculates confidence ellipses (for particular quantiles) for a scatter or group of points, using a parametric multivariate distribution e.g. multivariate normal. `Geom.ellipse` is `Geom.polygon` with `Stat.ellipse`, so I've expanded the `PolygonGeometry` type to work with such Statistics. 

```Julia
using RDatasets, Gadfly

D = dataset("datasets","faithful")
D[:g] = D[:Eruptions].>3.0

pa = plot(D, coord,
    x=:Eruptions, y=:Waiting, group=:g, Geom.point, Geom.ellipse
)
pb = plot(D, coord,
    x=:Eruptions, y=:Waiting, color=:g, Geom.point, Geom.ellipse,
    layer(Geom.ellipse(levels=[0.99]), style(line_style=:dot)),
    style(key_position=:none), Guide.ylabel(nothing)
)
hstack(pa,pb)
```
ellipse


bjarthur:
looks great!  thx for the tests and docs.

also, i noticed something wrong in a recent commit of yours.  the key title in [this example](http://gadflyjl.org/latest/lib/guides/guide_colorkey.html#Examples-1) is malformed.  could you please add a commit to this PR which fixes it?  thx.

bjarthur:
and don't worry about CI on 0.5 failing as [seemingly](https://github.com/JuliaLang/METADATA.jl/pull/13493) METADATA has dropped support for it.  if you wanted, you could add another commit to stop testing Gadfly on 0.5.

bjarthur:
a better fix would be to follow @tlnagy 's lead and make [render_colorkey_title](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/guide.jl#L372) look like [render(guide::Title...](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/guide.jl#L1011).  or would that mess up PNG output?

not sure we have a choice about dropping CI on 0.5.  but we should probably wait until 0.7 comes out to require 0.6.

Mattriks:
I can put a commit for my quick fix suggestion above in this PR for the colorkey title issue - if that's ok with you? A real fix should be done as a separate PR (ie to make multiline key titles work with svg).

bjarthur:
no, i don't like quick fixes.

Mattriks:
Ok, I'll open a separate PR sometime (or its fine if someone beats me to it!)

Mattriks:
There is a post over on discourse about travis, Julia 0.5 (and dropping support for 0.5): 
https://discourse.julialang.org/t/errors-for-git-pkg/9351/2

bjarthur:
thx again for this!

Issue #373, opened 2014-07-25
Multiple Axes


dcjones:
I'm going to create an issue for this and leave it open, since this comes up periodically, and I want to have conspicuous response.

I don't like plots with multiple axes. They're almost always terrible. Seriously, just do a google image search for “multiple axes”. You'll see some of the worst, most incomprehensible plots ever drawn. Stuff like this:

![multiple-y-axes-1](https://cloud.githubusercontent.com/assets/210361/3697737/b3f3c47c-13af-11e4-9201-dcdd93a4bea8.png)

That said, there are a handful of arguably legitimate uses, for example labeling a temperature scale in fahrenheit and celsius. So I'm not opposed to adding this feature, but there about a thousand more important things I need to do to make Gadfly great, and for me, this is very near the bottom of the list.

If someone wants to implement this, I can describe how to do it and review their PR, but I'm probably not going to do it myself within the next few years.

st33v3:
I generally agree that multiple scales (axes) are visually problematic. And I would be against adding scale to layers, but I would propose different approach. In some cases it still may be useful to present two "categories" in one axis, but just two, because a plot has two vertical and two horizontal edges.

For example - when data comes from stress testing of a web application. There is one dataset corresponding to number of virtual users visiting the application and other one corresponding to response time of the application. It usually happens that shape (or envelope) of response times follows number of virtual users. It is useful to display number of VU together with reply times. Obviously, VU number and time have different scale and units.

For that case I would recommend to introduce notion of secondary (minor) axes (both x and y) that have its Scale (Scale.x2_continuous, ...) and aesthetics - e.g.

    plot(dataset, x="Month",  y2="Approve", Scale.x2_continuous)

Labels and guides would be configured separately for x and x2 axis. I believe that this approach is more Gadfly like.

iq2luc:
Sometimes multiple Y axes (well... two) are important and very useful, for example looking at the attached plot one could easily understand if there is a capacitive or inductive load (source data is from a real capture, in this case there is a capacitve load because the current leads the voltage).
![vi-phase1](https://user-images.githubusercontent.com/13350093/50191903-a56dd000-0338-11e9-8076-e4c5b739885b.png)

ph-pi:
Double axis (and even more) are often used with time series. iq2luc give a pretty good example.
Financial time series also produces a lots of indicators zero based or significant when zero crossing.

At the moment, I use vstack, but it is not easy to compare precisely 2 or more plots.

[...]

Issue #77, opened 2013-09-19
Support non-Float64 number types


Keno:
It seems to me that sometimes Gadfly automatically scales the plot (e.g. loglog or something like that), but the y-axis label still says `f(x)`. When such an automatic transformation is applied and default labels are used, that should be fixed automatically. 

Relatedly, scale transforms currently expect the number to have a conversion to Float64. It would be nice to get rid of that as the rest of Gadfly is already blissfully ignorant to units, allowing things like 

```
plot(h->mgh,0Meter,12Meter)
```

using my SIUnits package. However, this doesn't work if a scale transform (even the identity transform) is being applied.

EDIT: Title updated to reflect real issue (see below).

PR #1105, opened 2018-03-16
Enable plotting of Unitful values

### LINKED ISSUE #1104: Gadfly can't plot Unitful.jl Quantities
e.g.:

```
julia> using Gadfly; using Unitful;
 
julia> plot(x=[1;2;3]u"m/s", y=[1;2;3]u"s", Geom.line)
Error showing value of type Gadfly.Plot:
ERROR: DimensionError: s and 1 are not dimensionally compatible.
Stacktrace:
 [1] optimize_ticks_typed(::Unitful.Quantity{Int64,Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)},Unitful.FreeUnits{(Unitful.Unit{:Second,Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)}}(0, 1//1),),Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)}}}, ::Unitful.Quantity{Int64,Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)},Unitful.FreeUnits{(Unitful.Unit{:Second,Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)}}(0, 1//1),),Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)}}}, ::Bool, ::Array{Tuple{Float64,Float64},1}, ::Int64, ::Int64, ::Int64, ::Float64, ::Float64, ::Float64, ::Float64, ::Bool) at /home/adam/.julia/v0.6/Gadfly/src/ticks.jl:72
 [2] (::Gadfly.#kw##optimize_ticks)(::Array{Any,1}, ::Gadfly.#optimize_ticks, ::Unitful.Quantity{Int64,Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)},Unitful.FreeUnits{(Unitful.Unit{:Second,Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)}}(0, 1//1),),Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)}}}, ::Unitful.Quantity{Int64,Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)},Unitful.FreeUnits{(Unitful.Unit{:Second,Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)}}(0, 1//1),),Unitful.Dimensions{(Unitful.Dimension{:Time}(1//1),)}}}) at ./<missing>:0
 [3] apply_statistic(::Gadfly.Stat.TickStatistic, ::Dict{Symbol,Gadfly.ScaleElement}, ::Gadfly.Coord.Cartesian, ::Gadfly.Aesthetics) at /home/adam/.julia/v0.6/Gadfly/src/statistics.jl:854
 [4] apply_statistics(::Array{Gadfly.StatisticElement,1}, ::Dict{Symbol,Gadfly.ScaleElement}, ::Gadfly.Coord.Cartesian, ::Gadfly.Aesthetics) at /home/adam/.julia/v0.6/Gadfly/src/statistics.jl:40
 [5] render_prepare(::Gadfly.Plot) at /home/adam/.julia/v0.6/Gadfly/src/Gadfly.jl:707
 [6] render(::Gadfly.Plot) at /home/adam/.julia/v0.6/Gadfly/src/Gadfly.jl:760
 [7] display(::Base.REPL.REPLDisplay{Base.REPL.LineEditREPL}, ::MIME{Symbol("text/html")}, ::Gadfly.Plot) at /home/adam/.julia/v0.6/Gadfly/src/Gadfly.jl:1074
 [8] display(::Base.REPL.REPLDisplay{Base.REPL.LineEditREPL}, ::Gadfly.Plot) at /home/adam/.julia/v0.6/Gadfly/src/Gadfly.jl:1019
 [9] display(::Gadfly.Plot) at ./multimedia.jl:194
 [10] eval(::Module, ::Any) at ./boot.jl:235
 [11] print_response(::Base.Terminals.TTYTerminal, ::Any, ::Void, ::Bool, ::Bool, ::Void) at ./REPL.jl:144
 [12] print_response(::Base.REPL.LineEditREPL, ::Any, ::Void, ::Bool, ::Bool) at ./REPL.jl:129
 [13] (::Base.REPL.#do_respond#16{Bool,Base.REPL.##26#36{Base.REPL.LineEditREPL,Base.REPL.REPLHistoryProvider},Base.REPL.LineEditREPL,Base.LineEdit.Prompt})(::Base.LineEdit.MIState, ::Base.AbstractIOBuffer{Array{UInt8,1}}, ::Bool) at ./REPL.jl:646
```

The problem I believe is here: https://github.com/GiovineItalia/Gadfly.jl/blob/aa6a1ad45e701a5af5a7761721dd617af8ee5c42/src/ticks.jl#L72

Instead of calling `convert(T, one(T))`, `oneunit(T)` should be called. May be other problems, but this is the first one.

non-Jedi:
Following test case works:

```
julia> using Gadfly; using Unitful;

julia> p = plot(x=[1;2;3;4]u"s", y=[2;4;8;16], Geom.line);

julia> draw(PNG("gadfly_unitful_test.png"), p);
```

Generates:

gadfly_unitful_test

Played around with it a bit afterward, and I didn't run into any other breaking issues. Would be nice if the ylabel showed units like the xlabel, but this at least makes Gadfly compatible with Unitful (and generally with well-behaved custom number types).

Closes #1104

### DIFF
--- a/src/geom/hvabline.jl
+++ b/src/geom/hvabline.jl
@@ -157,17 +157,15 @@ function render(geom::ABLineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetic
     # the line extends to the edges of the graph.
 
     if typeof(aes.y) <: Array{Function}
-        low, high = aes.xmin[1], aes.xmax[1]
+        lowx, highx = aes.xmin[1], aes.xmax[1]
     else
-        xextrema = extrema(aes.x)
-        yextrema = extrema(aes.y)
-        low = min(xextrema[1], yextrema[1])
-        high = max(xextrema[2], yextrema[2])
+        lowx, highx = extrema(aes.x)
     end
 
-    range = high-low
-    x0 = low-range
-    x1 = high+range
+    # extending the line to width 3 times x-range of data should be enough
+    rangex = highx - lowx
+    x0 = lowx - rangex
+    x1 = highx + rangex
 
     y0s = [x0 * m + b for (m,b) in zip(aes.slope, aes.intercept)]
     y1s = [x1 * m + b for (m,b) in zip(aes.slope, aes.intercept)]
--- a/src/scale.jl
+++ b/src/scale.jl
@@ -600,8 +600,7 @@ const color_continuous_gradient = color_continuous  ### WHY HAVE THIS ALIAS?
 
 function apply_scale(scale::ContinuousColorScale,
                      aess::Vector{Gadfly.Aesthetics}, datas::Gadfly.Data...)
-    cdata = reduce(vcat, [], [i.color for i in datas if i.color != nothing])
-    filter!(!isna, cdata)
+    cdata = skipmissing(Iterators.flatten(i.color for i in datas if i.color != nothing))
     if !isempty(cdata)
       cmin, cmax = extrema(cdata)
     else
--- a/src/scale.jl
+++ b/src/scale.jl
@@ -600,25 +600,13 @@ const color_continuous_gradient = color_continuous  ### WHY HAVE THIS ALIAS?
 
 function apply_scale(scale::ContinuousColorScale,
                      aess::Vector{Gadfly.Aesthetics}, datas::Gadfly.Data...)
-    cmin, cmax = Inf, -Inf
-    for data in datas
-        data.color === nothing && continue
-
-        for c in data.color
-            ismissing(c) && continue
-
-            c = convert(Float64, c)
-            if c < cmin
-                cmin = c
-            end
-
-            if c > cmax
-                cmax = c
-            end
-        end
+    cdata = reduce(vcat, [], [i.color for i in datas if i.color != nothing])
+    filter!(!isna, cdata)
+    if !isempty(cdata)
+      cmin, cmax = extrema(cdata)
+    else
+        return
     end
-
-    (cmin == Inf || cmax == -Inf) && return
 
     if scale.minvalue != nothing
         cmin = scale.minvalue
@@ -671,11 +659,11 @@ function apply_scale(scale::ContinuousColorScale,
 end
 
 function apply_scale_typed!(ds, field, scale::ContinuousColorScale,
-                            cmin::Float64, cspan::Float64)
+                            cmin, cspan)
     for (i, d) in enumerate(field)
         if isconcrete(d)
             ds[i] = convert(RGB{Float32},
-                        scale.f((convert(Float64, scale.trans.f(d)) - cmin) / cspan))
+                        scale.f((scale.trans.f(d) - cmin) / cspan))
         else
             ds[i] = missing
         end
--- a/src/ticks.jl
+++ b/src/ticks.jl
@@ -2,7 +2,7 @@
 # little opaque because I want to avoid assuming the log function is defined
 # over typeof(xspan)
 function bounding_order_of_magnitude(xspan::DT) where DT
-    one_dt = convert(DT, one(DT))
+    one_dt = oneunit(DT)
 
     a = 1
     step = 1
@@ -69,7 +69,7 @@ function optimize_ticks_typed(x_min::T, x_max::T, extend_ticks,
                            granularity_weight::Float64, simplicity_weight::Float64,
                            coverage_weight::Float64, niceness_weight::Float64,
                            strict_span) where T
-    one_t = convert(T, one(T))
+    one_t = oneunit(T)
     if x_max - x_min < eps()*one_t
         R = typeof(1.0 * one_t)
         return R[x_min], x_min - one_t, x_min + one_t
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -2,3 +2,4 @@ DataFrames
 RDatasets
 Cairo
 CSV
+Unitful
--- a/test/testscripts/percent.jl
+++ b/test/testscripts/percent.jl
@@ -2,7 +2,7 @@ using Gadfly
 
 set_default_plot_size(6inch, 6inch)
 
-import Base: show, +, -, /, *, isless, one, zero, isfinite
+import Base: convert, show, +, -, /, *, isless, one, zero, isfinite
 
 struct Percent
     value::Float64
@@ -23,6 +23,7 @@ isless(a::Percent, b::Percent) = isless(a.value, b.value)
 one(::Type{Percent}) = Percent(0.01)
 zero(::Type{Percent}) = Percent(0.0)
 isfinite(a::Percent) = isfinite(a.value)
+convert(::Type{Float64}, x::Percent) = x.value
 show(io::IO, p::Percent) = print(io, round(100 * p.value, digits=4), "%")
 
 y=[Percent(0.1), Percent(0.2), Percent(0.3)]
--- a/test/testscripts/unitful_basic.jl
+++ b/test/testscripts/unitful_basic.jl
@@ -1,0 +1,19 @@
+using Unitful, Gadfly, DataFrames
+
+a = -9.81u"m/s^2"
+t = (1:0.5:10)u"s"
+v = a * t
+v2 = v + 2u"m/s"
+h = 0.5 * a * t.^2
+df = DataFrame(time=t, velocity=v, position=h,
+               unitlesst=ustrip.(t), unitlessv=ustrip.(v), unitlessh=ustrip.(h),
+               position2=reverse(h))
+
+# test basics of point/line plots with Unitful
+gridstack(reshape([plot(df, x=:time, y=:velocity),
+                   plot(df, x=:position, y=:time),
+                   plot(df, x=:unitlesst, y=:velocity),
+                   plot(df, x=:time, y=:unitlessh),
+                   plot(df, x=:time, y=:position, Geom.line),
+                   plot(df, layer(x=:time, y=:position, Geom.line),
+                        layer(x=:time, y=:position2))], (3,2)))
--- a/test/testscripts/unitful_basic.jl
+++ b/test/testscripts/unitful_basic.jl
@@ -2,9 +2,9 @@ using Unitful, Gadfly, DataFrames
 
 a = -9.81u"m/s^2"
 t = (1:0.5:10)u"s"
-v = a * t
-v2 = v + 2u"m/s"
-h = 0.5 * a * t.^2
+v = a .* t
+v2 = v .+ 2u"m/s"
+h = 0.5 * a .* t.^2
 df = DataFrame(time=t, velocity=v, position=h,
                unitlesst=ustrip.(t), unitlessv=ustrip.(v), unitlessh=ustrip.(h),
                position2=reverse(h))
--- a/test/testscripts/unitful_color.jl
+++ b/test/testscripts/unitful_color.jl
@@ -2,9 +2,9 @@ using Unitful, Gadfly, DataFrames
 
 a = -9.81u"m/s^2"
 t = (1:0.5:10)u"s"
-v = a * t
-v2 = v + 2u"m/s"
-h = 0.5 * a * t.^2
+v = a .* t
+v2 = v .+ 2u"m/s"
+h = 0.5 * a .* t.^2
 df = DataFrame(time=t, velocity=v, position=h,
                unitlesst=ustrip.(t), unitlessv=ustrip.(v), unitlessh=ustrip.(h),
                position2=reverse(h))
--- a/test/testscripts/unitful_color.jl
+++ b/test/testscripts/unitful_color.jl
@@ -1,0 +1,15 @@
+using Unitful, Gadfly, DataFrames
+
+a = -9.81u"m/s^2"
+t = (1:0.5:10)u"s"
+v = a * t
+v2 = v + 2u"m/s"
+h = 0.5 * a * t.^2
+df = DataFrame(time=t, velocity=v, position=h,
+               unitlesst=ustrip.(t), unitlessv=ustrip.(v), unitlessh=ustrip.(h),
+               position2=reverse(h))
+
+# Test that it's possible to categorize by Unitful quantities
+vstack(plot(df, x=:time, y=:position, color=:velocity, Geom.point),
+       plot(df, x=:time, y=:velocity, color=:position, Geom.line),
+       plot(df, x=:time, y=:position, color=:unitlessv))
--- a/test/testscripts/unitful_geoms.jl
+++ b/test/testscripts/unitful_geoms.jl
@@ -1,0 +1,26 @@
+using Unitful, Gadfly, DataFrames
+
+a = -9.81u"m/s^2"
+t = (1:0.5:10)u"s"
+v = a * t
+v2 = v + 2u"m/s"
+h = 0.5 * a * t.^2
+df = DataFrame(time=t, velocity=v, position=h,
+               unitlesst=ustrip.(t), unitlessv=ustrip.(v), unitlessh=ustrip.(h),
+               position2=reverse(h))
+
+# Test various geometries with Unitful
+p1 = plot(df, x=:time, y=:velocity, Geom.bar)
+p2 = plot(df, x=:time, y=:position, Geom.point,
+          intercept=[-80u"m"], slope=[10u"m/s"], Geom.abline)
+p3 = plot(df, x=:time, y=:velocity, Geom.line,
+          yintercept=[-20u"m/s", -40u"m/s"], Geom.hline)
+# Currently explicitly stated that it loess and lm require arrays of plain numbers
+#p4 = plot(df, x=:time, y=:position, Geom.point,
+#          Geom.smooth(method=:loess, smoothing=0.2))
+p4 = plot(df, x=:position, y=:velocity, Geom.path)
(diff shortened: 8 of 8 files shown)


non-Jedi:
@bjarthur do you need me to include the fixes to the y-axis label with this PR?
This is a relatively simple bugfix, but I wouldn't know really where to start on
that one. I plan on fixing that piece eventually, but it may be awhile...

bjarthur:
by fixing the y-axis label, to you actually mean y-axis ticks?  if so, the relevant code is [here](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/guide.jl#L735).  compare with the x-axis tick code immediately above to see what the difference might be.

also, if we're serious about supporting this in future, you should add your simple test case above to test/testscripts/.  this would necessitate adding Unitful to test/REQUIRE.

@tlnagy what do you think of adding a test dependency like this?  previously you had mentioned refactoring Measures out in favor of Unitful i believe.

tlnagy:
> what do you think of adding a test dependency like this? previously you had mentioned refactoring Measures out in favor of Unitful i believe.

I don't have any problems with adding a test dependency since that wouldn't bloat our dependency graph for a normal install.

We should definitely add a test for using `Unitful.jl` with Gadfly in this PR to make sure that we don't break compatibility in the future. 

> Would be nice if the ylabel showed units like the xlabel

You're not testing unitful quantities on the y axis though? `y=[2;4;8;16]` are Ints not Unitful quantities?

@non-Jedi Some other edge cases: does this work with unitful quantities on the y axis? On both axes?

tlnagy:
> do you need me to include the fixes to the y-axis label with this PR?

We should make sure that this works in all basic cases before merging this PR. We can leave this open for one of us (or someone else) to update if you don't have time. You can keep using your local branch with this modification for your immediate plotting needs.

tlnagy:
Checking out this PR seems to work for the y axis too?

```julia
p = plot(x=[1;2;3;4]u"s", y=[2;4;8;16]u"kg", Geom.line)
```

image

tlnagy:
One thing I would like to point out is that it is custom to have the units in axes labels instead in the ticks. Like so

```julia
p = plot(x=[1;2;3;4], y=[2;4;8;16]u"kg", Geom.line, Guide.ylabel("y (kg)"))
```
image

but here the information is duplicated and there is no easy way to hide the units in the ticks

```julia
p = plot(x=[1;2;3;4], y=[2;4;8;16]u"kg", Geom.line, Guide.yticks(ticks=[2,4,6,8]))

DimensionError: 2 and 2 kg are not dimensionally compatible.

Stacktrace:
 [1] min(::Unitful.Quantity{Int64,Unitful.Dimensions{(Unitful.Dimension{:Mass}(1//1),)},Unitful.FreeUnits{(Unitful.Unit{:Gram,Unitful.Dimensions{(Unitful.Dimension{:Mass}(1//1),)}}(3, 1//1),),Unitful.Dimensions{(Unitful.Dimension{:Mass}(1//1),)}}}, ::Int64) at ./operators.jl:361
 [2] apply_statistic(::Gadfly.Stat.TickStatistic, ::Dict{Symbol,Gadfly.ScaleElement}, ::Gadfly.Coord.Cartesian, ::Gadfly.Aesthetics) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/statistics.jl:784
 [3] apply_statistics(::Array{Gadfly.StatisticElement,1}, ::Dict{Symbol,Gadfly.ScaleElement}, ::Gadfly.Coord.Cartesian, ::Gadfly.Aesthetics) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/statistics.jl:41
 [4] render_prepare(::Gadfly.Plot) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/Gadfly.jl:708
 [5] render(::Gadfly.Plot) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/Gadfly.jl:761
 [6] show at /Users/tamasnagy/.julia/v0.6/Gadfly/src/Gadfly.jl:965 [inlined]
 [7] limitstringmime(::MIME{Symbol("image/svg+xml")}, ::Gadfly.Plot) at /Users/tamasnagy/.julia/v0.6/IJulia/src/inline.jl:24
 [8] display_dict(::Gadfly.Plot) at /Users/tamasnagy/.julia/v0.6/IJulia/src/execute_request.jl:33
 [9] (::Compat.#inner#17{Array{Any,1},IJulia.#display_dict,Tuple{Gadfly.Plot}})() at /Users/tamasnagy/.julia/v0.6/Compat/src/Compat.jl:488
 [10] execute_request(::ZMQ.Socket, ::IJulia.Msg) at /Users/tamasnagy/.julia/v0.6/IJulia/src/execute_request.jl:186
 [11] (::Compat.#inner#17{Array{Any,1},IJulia.#execute_request,Tuple{ZMQ.Socket,IJulia.Msg}})() at /Users/tamasnagy/.julia/v0.6/Compat/src/Compat.jl:488
 [12] eventloop(::ZMQ.Socket) at /Users/tamasnagy/.julia/v0.6/IJulia/src/eventloop.jl:8
 [13] (::IJulia.##14#17)() at ./task.jl:335
```

non-Jedi:
>You're not testing unitful quantities on the y axis though? y=[2;4;8;16] are
>Ints not Unitful quantities?

🤦 yep, that explains it.

I'll plan on pushing a commit to this branch sometime this weekend testing a
bunch of potential codepaths for Unitful use. No promises after that.

To be honest, I see the current content of this PR as a bugfix with the topic of
things like turning off display of units as ticks as enhancements.

non-Jedi:
Interesting note: using Unitful here enforces the no-multiple axes principle
articulated in https://github.com/GiovineItalia/Gadfly.jl/issues/373 much more
strictly. It also may provide a clean api for secondary axes if anyone wants to
implement that. since it locks the scale of the axis to the type of the
displayed data, you could have secondary axes generated based on mismatches in
type of y-values in different layers.

non-Jedi:
@bjarthur @tlnagy FYI, from the perspective of getting all basic Unitful
plotting functionality working, I believe this PR is complete now. Still some
enhancements to be made wrt things like putting units in axis labels instead of
tick labels as discussed above, but from my perspective that could be done in a
later PR (which I'm planning on doing... eventually...). Your call, of course,
whether to merge now or wait for those enhancements.

src/scale.jl
```diff
@@ -575,26 +575,14 @@ element_aesthetics(::ContinuousColorScale) = [:color]
...
+    cdata = reduce(vcat, [], [i.color for i in datas if i.color != nothing])
+    filter!(!isna, cdata)
+    if !isempty(cdata)
+      cmin, cmax = extrema(cdata)
+    else
+        return
     end
 
-    (cmin == Inf || cmax == -Inf) && return
-
```
bjarthur: do the changes above do the exact same thing as before?  or do they have a material bearing on making Unitful work?  if the former, have you by chance tested for speed?  frequently hand made for-loops in Julia are faster than the convenience functions.

src/geom/hvabline.jl
```diff
@@ -138,17 +138,15 @@ function render(geom::ABLineGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetic
...
+        lowx, highx = extrema(aes.x)
     end
 
-    range = high-low
-    x0 = low-range
-    x1 = high+range
+    # extending the line to width 3 times x-range of data should be enough
+    rangex = highx - lowx
+    x0 = lowx - rangex
+    x1 = highx + rangex
```
bjarthur: what do these changes have to do with making Unitful work?  have you done regression testing to make sure nothing has changed by omitting the y-axis extremes?

non-Jedi: @bjarthur you can't compare y with x if y and x have different units. I regression tested, and the only differences I saw on my machine were between .js.svg graphs, which is to be expected as I understand it.

src/scale.jl
non-Jedi: This does the same thing (while allowing for more types of numbers of course). I haven't benchmarked yet, but I'll do so.  Hadn't thought this method would have a large performance impact, but I see now that there's every potential it could since it loops through the entire color aesthetic at least once.

While I'm digging in, I'll see if it's possible to make this method type-stable.

src/geom/hvabline.jl
non-Jedi: @bjarthur you can't compare y with x if y and x have different units. I regression tested, and the only differences I saw on my machine were between .js.svg graphs, which is to be expected as I understand it.

src/scale.jl
non-Jedi: This does the same thing (while allowing for more types of numbers of course). I haven't benchmarked yet, but I'll do so. Hadn't thought this method would have a large performance impact, but I see now that there's every potential it could since it loops through the entire color aesthetic at least once.

While I'm digging in, I'll see if it's possible to make this method type-stable.

non-Jedi: Ran a little mini-diagnostic where I encapsulated the changed functionality into
a local function and called it. Example for the new version:

```julia
    function internal_test(datas::Gadfly.Data...)
        cdata = reduce(vcat, [], [i.color for i in datas if i.color != nothing])
        filter!(!isna, cdata)
        if !isempty(cdata)
          cmin, cmax = extrema(cdata)
        else
            return
        end
    end
    @time cmin, cmax = internal_test(datas...)
```

(Yes, yes I know it's not correct, but it's correct for the happy path, and
that's enough for this benchmark). Results for the new version:

```
julia> using Gadfly;
julia> x = 1:10000;
julia> y = rand(1:100, 10000);
julia> c = rand(1:250, 10000);
julia> plot(x=x, y=y, color=c)
  0.212279 seconds (93.11 k allocations: 5.061 MiB)
  0.000835 seconds (22 allocations: 78.953 KiB)
  0.000785 seconds (22 allocations: 78.953 KiB)
julia> plot(x=x, y=y, color=c)
  0.001313 seconds (22 allocations: 78.953 KiB)
  0.000730 seconds (22 allocations: 78.953 KiB)
  0.000760 seconds (22 allocations: 78.953 KiB)
```

Modified code for old version:

```julia
    function internal_test(datas::Gadfly.Data...)
        cmin, cmax = Inf, -Inf
        for data in datas
            data.color === nothing && continue

            for c in data.color
                c === NA && continue

                c = convert(Float64, c)
                if c < cmin
                    cmin = c
                end

                if c > cmax
                    cmax = c
                end
            end
        end

        (cmin == Inf || cmax == -Inf) && return
        cmin, cmax
    end
    @time cmin, cmax = internal_test(datas...)
```

Results for old version:

```
julia> using Gadfly;
julia> x = 1:10000;
julia> y = rand(1:100, 10000);
julia> c = rand(1:250, 10000);
julia> plot(x=x, y=y, color=c)
  0.023244 seconds (33.02 k allocations: 805.858 KiB)
  0.002329 seconds (29.50 k allocations: 617.172 KiB)
  0.002437 seconds (29.50 k allocations: 617.172 KiB)
julia> plot(x=x, y=y, color=c)
  0.002203 seconds (29.50 k allocations: 617.172 KiB)
  0.002188 seconds (29.50 k allocations: 617.172 KiB)
  0.002348 seconds (29.50 k allocations: 617.172 KiB)
```

Believe it or not, this change is actually a performance improvement if
anything. Looks like the compiler can basically optimize out the new stuff.

bjarthur:
the changes here are small, but the implications are much bigger i think.  i need more time to understand the impact of having non-Float64 aesthetics.  @tlnagy @Mattriks thoughts?

non-Jedi:
FWIW, it was @dcjones intention for non-Float64 to work back in 2013: https://github.com/GiovineItalia/Gadfly.jl/issues/77

bjarthur:
ahah, so nominally unitful should've worked all along, and the changes you had to make in this PR were simply to fix oversights due to lack of unit testing combining Percent with hvabline, scales, ticks, etc.

okay, so i'm fine then with merging.

my only reservation is adding Unitful to test/REQUIRE, because 1. there are other packages which also provide unit types; why should we single out that one?, and 2. Gadfly already has a huge dependency tree which many complain about.

would you consider adding a gridstack to test/percent.jl and moving your three test scripts over there?  you also need to resolve conflicts.

thanks for this!  looks great.

tlnagy:
> my only reservation is adding Unitful to test/REQUIRE, because 1. there are other packages which also provide unit types; why should we single out that one?

I think `Unitful` is the defactor standard at this point. I'm okay with adding it.

> 2. Gadfly already has a huge dependency tree which many complain about.

This is only in `test/REQUIRE` not in `REQUIRE`. This would only bloat things if you want to run tests locally. Assuming we're doing our job properly, users shouldn't need to do that so I'm okay with adding it.

Main things remaining before this is ready:

- [ ] Doc examples
- [ ] Add to NEWS.md

tlnagy:
> it is now, but for how long? there have been many others in the past. i'd rather gadfly be agnostic.

If a new one emerges, we can always switch the tests over to the new library. None of the changes this PR makes in `src/` are specific to `Unitful` and we can always switch libraries in `test/` without breaking anything.

bjarthur:
do bug fixes like this warrant a mention in NEWS?  i'd suggest holding off on that until the units appear in the label and not the ticks.  that would be a big enough step that a whole new section in the Tutorial dedicated to units could be added too.

also, i am still really in favor of just having the unit tests in this PR use the Percent type that dcjones already has in the tests.  the code in the tests as they are now could be saved and used for the docs later.

lastly, worth nothing that when i cut and paste the current tests into the REPL i get a `DimensionError`.

tlnagy:
> do bug fixes like this warrant a mention in NEWS?

I was thinking of it more like we're adding feature (e.g. Gadfly now supports unitful values too!)

> also, i am still really in favor of just having the unit tests in this PR use the Percent type that dcjones already has in the tests. the code in the tests as they are now could be saved and used for the docs later.

If you feel strongly about this then we shouldn't include `Unitful` in the tests. You're talking about this, right?

https://github.com/GiovineItalia/Gadfly.jl/blob/602f2d6e7722a06d8bdffe896baecb809bd8bf52/test/testscripts/percent.jl#L7-L9

bjarthur:
as @non-Jedi  [pointed out](https://github.com/GiovineItalia/Gadfly.jl/pull/1105#issuecomment-377709990), units have nominally worked all the way back to #77.  they just broke over time because of lack of testing.  so yes, let's just expand testscripts/percent.jl to test everything that this PR fixes.  then we don't need Unitful in test/REQUIRE and Gadfly code and tests are agnostic.  i don't mind using Unitful in the docs, but think it looks funny to have the units in the ticks not labels, so would wait to do so until that is changed.

non-Jedi:
On the subject of units on ticks vs labels, it looks like nicely formatting this
kind of thing is exactly what https://github.com/dcjones/Showoff.jl is for,
which appears to still be a dependency of Gadfly. I haven't messed with it
myself, but getting Unitful values to show on ticks with just the bare numbers
should just be a matter of defining
`Showoff.showoff(::AbstractArray{Unitful.Quantity})`. So even that functionality
might just be a matter of documenting stuff better, though I haven't messed with
Showoff myself to check.

That said, if there's consensus on not testing against Unitful, I'll go ahead
and migrate all the new tests into the percent testscript as best I can. And
then rebase on master of course. Will do as soon as I get a chance.

tlnagy:
What's wrong with where it is?

```julia
julia> Pkg.dependents("Showoff")
4-element Array{AbstractString,1}:
 "Plots"
 "Gadfly"
 "Mamba"
 "MixedModels"
```

If you want to make a PR against Showoff adding that convenience function, I would be happy to review it and merge. 

Thanks for your efforts here!

non-Jedi:
No way of doing it without making Showoff depend on Unitful which would then end up pulling in Unitful to Gadfly where we're trying to avoid it. :/ Unless y'all know how to do some sort of an optional dependency, I was thinking that I would just write some more thorough documentation about overriding `Showoff.showoff`.

bjarthur:
optional dependencies will soon be added to Pkg3.  i look forward to making Gadfly optionally depend on Cairo and DataFrames.

in this case though, an example in the documentation of how to use Showoff with Unitful would be great.  i'd even devote an entirely new section in the Tutorial or Manual to it.  could do that now and not wait until we've figured out how to automatically add it to labels.

non-Jedi:
For what it's worth, this is now good to go with Julia 1.0. Only thing blocking merge now is the request to remove the test dependency on Unitful. I hope to have time to clean that up sometime in the next month or so, but no promises.

non-Jedi:
I'm not going to be able to test this effectively without Unitful. As an example, take a look at https://github.com/GiovineItalia/Gadfly.jl/pull/1105/files#diff-d568a2b9faaba434884de694f9537f19R15 where I'm plotting distance versus time and giving slope in terms of velocity; that's too much of Unitful to reimplement just for running a test. @bjarthur, could you please reconsider allowing Unitful as a test dependency?

bjarthur:
thanks for this!  would be great to some day document how to use Showoff to get the units in the labels and out of the ticks.

PR #1106, opened 2018-03-16
grid_line_style

Mattriks:
This issue came up in this discourse [post](https://discourse.julialang.org/t/ggplot-style-background-in-gadfly/9722). I've now changed `Theme(grid_strokedash= )` to `Theme(grid_line_style= )`. The behaviour of `grid_line_style` matches the behaviour of `line_style` in Theme, plus I edited the Theme doc. 

```Julia
using Colors, RDatasets, Gadfly
palette(n::Int) = LCHuv.(65, 100, 15:(360/n):374)
colscale = Scale.color_discrete(palette)
ggplot_theme = Theme(panel_fill="gray90", grid_color="white", grid_line_style=:solid, discrete_color_scale=colscale)

p = plot(dataset("datasets","iris"), x=:PetalWidth, y=:SepalLength, color=:Species, Geom.point, ggplot_theme)
```
themes

Also, Gadfly currently has 2 named Themes: `:default` and `:dark`. Would anybody object to having another optional (not default) named Theme: `:ggplot`?

### DIFF
--- a/docs/src/man/themes.md
+++ b/docs/src/man/themes.md
@@ -44,6 +44,7 @@ These parameters can either be used with `Theme` or `style`
   * `grid_color_focused`: In the D3 backend, mousing over the plot makes the
     grid lines emphasised by transitioning to this color. (Color or Nothing)
   * `grid_line_width`: Width of grid lines. (Measure)
+  * `grid_line_style`: Style of grid lines. (Symbol in :solid, :dash, :dot, :dashdot, :dashdotdot, or Vector of Measures)   
   * `minor_label_font`: Font used for minor labels such as tick labels and entries
     in keys. (String)
   * `minor_label_font_size`: Font size used for minor labels. (Measure)
@@ -87,8 +88,8 @@ These parameters can either be used with `Theme` or `style`
   * `bar_highlight`: Color used to stroke bars in bar plots. If a function is
     given, it's used to transform the fill color of the bars to obtain a stroke
     color. (Function, Color, or Nothing)
-  * `discrete_color_scheme`: A `DiscreteColorScale` see [Scale.color_discrete_hue](@ref)
-  * `continuous_color_scheme`: A `ContinuousColorScale` see [Scale.color_continuous](@ref)
+  * `discrete_color_scale`: A `DiscreteColorScale` see [Scale.color_discrete_hue](@ref)
+  * `continuous_color_scale`: A `ContinuousColorScale` see [Scale.color_continuous](@ref)
 
 ## Examples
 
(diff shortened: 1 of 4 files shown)


tlnagy:
I think that looks great! It would be great to have a `ggplot` style. Thanks @Mattriks!

tlnagy:
Also, @Mattriks it might be nice to add the ggplot theme to the themes page http://gadflyjl.org/latest/man/themes.html and maybe even feature it in the tutorial for exposure.

Mattriks:
I started looking at implementing a named theme `:ggplot`, but ran into an issue:

```Julia
using Colors, RDatasets

dpalette(n::Int) = LCHuv.(65, 100, 15:(360/n):374)

# This is a new DiscreteColorScale for the ggplot theme:
function ggplot_color_discrete(;levels=nothing, order=nothing, preserve_order=true)
    Gadfly.Scale.DiscreteColorScale(dpalette, levels=levels, order=order, preserve_order=preserve_order)
end

ggplot_theme = Theme(panel_fill="gray92", 
    grid_color="white", 
    grid_line_style=:solid, 
    discrete_color_scale=ggplot_color_discrete()
)

iris = dataset("datasets","iris")
pa = plot(iris, x=:PetalWidth, y=:SepalLength, color=:Species, Geom.point, ggplot_theme,
    Scale.color_discrete(order=[3,2,1]),
    Guide.colorkey(title="Iris", pos=[0.1, 9.3]) )

pb = plot(iris, x=:PetalWidth, y=:SepalLength, color=:Species, Geom.point, ggplot_theme,
    ggplot_color_discrete(order=[3,2,1]),
    Guide.colorkey(title="Iris", pos=[0.1, 9.3]) )
hstack(pa, pb)
```
ggplot_theme

Both plots use `ggplot_theme`. In the left plot, I attempt to change the order of the key colors with `Scale.color_discrete`, but `Scale.color_discrete` and `ggplot_color_discrete` are not the same color scale.  In order to change the order of the key colors, I have to use the Theme-specific `DiscreteColorScale`, which is what I do in the right plot.  The issue here is that currently, `Theme` wants you to replace the whole color scale, so you have to write a whole new color scale for each named theme (this is currently what the :dark theme does).  Also try changing the order of the key colors using the :dark theme.

The more efficient way to do this would be for `Theme` to replace just the colormap function. Note that for a `DiscreteColorScale` the colormap function is called `f` (and for a `ContinuousColorScale` its called `colormap`). Letting `Theme` control the colormap functions (for `DiscreteColorScale` and `ContinuousColorScale`), would then mean that users would be able to control the Scale levels, order, minvalue, maxvalue etc just using `Scale.color_discrete` and `Scale.color_continuous`.  Hence it seems a bigger PR is required here, which might take some time.

Issue #1112, opened 2018-03-17
fix colorkey example


bjarthur:
the legend title is malformed [here](http://gadflyjl.org/latest/lib/guides/guide_colorkey.html#Examples-1).  it's supposed to be two lines but it's not.  i can't get multi-line text to work anywhere in SVG.  it that even possible?  if not, we should just change the example to use a single line.

tlnagy:
Maybe @Mattriks knows?

bjarthur:
fixed by https://github.com/GiovineItalia/Gadfly.jl/pull/1124

Issue #766, opened 2015-12-09
Feature request: map multiple numerical and categorical data to plots


ljournot:
I am 1) a newbie to Julia (and to programming) 2) a biologist 3) French. It's tough but c'est la vie !
In biology we often want to map several kind of numerical (e.g. fold enrichment, gene counts, p-value) and categorical (e.g. gene ontology category) data in a single plot. Some colleagues of mine produced the attached plot using R and ggplot2. I failed to produce a similar plot using Gadfly. I could map the p-value to the point color (Scale.ContinuousColorScale(Scale.lab_gradient(colorant"white",colorant"orange",colorant"red"))), but could not change their shape and size to map GO category and gene count, respectively. I suspect a solution might be something like change the Theme for each point, but it sounds quite complicated and I still don't find out how to change the shape depending on categorical values. Any idea ?

Thanks a lot for your help.

laurent

![fairy_lights_go](https://cloud.githubusercontent.com/assets/16228350/11693123/66f8e528-9ea3-11e5-9cd1-1261f670f851.png)

tbreloff:
I might be able to help, but it would be much easier if you supplied your data.  Can you copy it into into a github gist?

tbreloff:
So I took a stab at this using Plots.jl (honestly, I don't plan on attempting this in pure Gadfly, I think it will be too annoying)

![tmp](https://cloud.githubusercontent.com/assets/933338/11695564/e15fa5de-9e7d-11e5-9451-760a4c83c695.png)

Feel free to open issues at https://github.com/tbreloff/Plots.jl/issues with questions/requests.

ljournot:
Hi Tom. Thanks for taking care of my request. I will look at Plots over the next few days. Just in case you want to go on, please find attached our data. laurent

[Data.txt](https://github.com/dcjones/Gadfly.jl/files/58366/Data.txt)

ljournot:
it looks like the Gadfly backend used by Plots has an intrinsic problem with variable marker size.

Here is my script:
using DataFrames
cd("C:/Users/Laurent/Documents/LAURENT/MANIPS/Julia scripts/FairyLights")
data=readtable("Input//Data.csv");
using Plots
gadfly()
dataframes()
plot(data,:Fold_Enrichment,:Term,group=:Category,size=(500,500),
[...]

Issue #602, opened 2015-06-03
Find out what the default Gadfly color values are


woest:
I know I can introduce colors manually, but I like the color scheme of Gadfly, and could not find an easy way to retrieve a vector of colors in the default order as used by Gadfly.

Did I miss something? I could not find anything in the documentation, and the closest I got (but did not know what to do with it) by browsing through the code is Gadfly.Scale.color_discrete

dcjones:
The color palette is generated, so you won't find all the colors listed explicitly anywhere. But you can generate them yourself. Here's the code Gadfly uses:

```julia
using Color

n = 12
distinguishable_colors(n, ColorValue[LCHab(70, 60, 240)],
                       transform=c -> deuteranopic(c, 0.5),
                       lchoices=Float64[65, 70, 75, 80],
                       cchoices=Float64[0, 50, 60, 70],
                       hchoices=linspace(0, 330, 24))
```

You can change `n` to get the desired number of colors (which can be arbitrarily large).

woest:
Nice :). Thanks a lot for that. 

Possibly a nice addition for the documentation? 

Or even add a wrapper function around it so people can call it with always the same lchoices/cchoices/hchoices Gadfly uses (if those settings ever change in the future).

woest:
Nevermind, I should have known how to do this:

```julia
julia> using Gadfly

julia> a = Scale.color_discrete_hue()
DiscreteColorScale((anonymous function),nothing,nothing,true)

julia> a.f(3)
3-element Array{ColorValue{T},1}:
 LCHab{Float64}(70.0,60.0,240.0)             
 LCHab{Float64}(80.0,70.0,100.43478260869566)
 LCHab{Float64}(65.0,70.0,0.0)    
```

GordStephen:
@woest Great tip, thanks!

PR #1116, opened 2018-03-17
refactor docs to use docstrings and have a gallery

bjarthur:
so two things that'd be nice to have are:

1. a [matplotlib style gallery](https://matplotlib.org/gallery/index.html), where example plots are all collected in one place along with the exact code to reproduce them.

2. julia style docstrings so that documentation can be accessed on the REPL.  docs embedded in the code are also less likely to get out of date.

towards these ends, i've started down the path of refactoring the documentation.  it's a big change though, so this PR currently only contains 10% or so of the changes needed, as i want to get feedback before proceeding.

would be great if folks could checkout this PR, do a `julia docs/make.jl`, then open `docs/build/index.html` and have a look around.  also do a `?Geom`, `?Geom.bar`, etc. on the REPL.  i've only made galleries for, and altered the libraries of, two Geoms and one Guide.  the rest will follow suit if you guys like this new format.  let me know!  thanks.


tlnagy:
Doesn't build for me?

```julia
$ julia docs/make.jl
Documenter: setting up build directory.
Documenter: expanding markdown templates.
ERROR: LoadError: MethodError: no method matching discretize_make_pda(::CategoricalArrays.CategoricalArray{String,1,UInt8,String,CategoricalArrays.CategoricalString{UInt8},Union{}}, ::Array{CategoricalArrays.CategoricalString{UInt8},1})
Closest candidates are:
  discretize_make_pda(!Matched::Array{T,1} where T, ::Any) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/scale.jl:263
  discretize_make_pda(!Matched::DataArrays.DataArray, ::Any) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/scale.jl:271
  discretize_make_pda(!Matched::Range, ::Any) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/scale.jl:279
  ...
Stacktrace:
 [1] discretize(::CategoricalArrays.CategoricalArray{String,1,UInt8,String,CategoricalArrays.CategoricalString{UInt8},Union{}}, ::Void, ::Void, ::Bool) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/scale.jl:301
 [2] discretize(::CategoricalArrays.CategoricalArray{String,1,UInt8,String,CategoricalArrays.CategoricalString{UInt8},Union{}}, ::Void, ::Void) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/scale.jl:295
 [3] apply_scale(::Gadfly.Scale.DiscreteScale, ::Array{Gadfly.Aesthetics,1}, ::Gadfly.Data, ::Vararg{Gadfly.Data,N} where N) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/scale.jl:366
 [4] apply_scales(::IterTools.Distinct{Base.ValueIterator{Dict{Symbol,Gadfly.ScaleElement}},Gadfly.ScaleElement}, ::Array{Gadfly.Aesthetics,1}, ::Gadfly.Data, ::Vararg{Gadfly.Data,N} where N) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/scale.jl:33
 [5] apply_scales(::IterTools.Distinct{Base.ValueIterator{Dict{Symbol,Gadfly.ScaleElement}},Gadfly.ScaleElement}, ::Gadfly.Data) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/scale.jl:52
 [6] render_prepare(::Gadfly.Plot) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/Gadfly.jl:674
 [7] render(::Gadfly.Plot) at /Users/tamasnagy/.julia/v0.6/Gadfly/src/Gadfly.jl:752
 [8] show at /Users/tamasnagy/.julia/v0.6/Gadfly/src/Gadfly.jl:956 [inlined]
 [9] verbose_show(::Base.AbstractIOBuffer{Array{UInt8,1}}, ::MIME{Symbol("image/svg+xml")}, ::Gadfly.Plot) at ./multimedia.jl:42
 [10] #sprint#228(::Void, ::Function, ::Int64, ::Function, ::MIME{Symbol("image/svg+xml")}, ::Vararg{Any,N} where N) at ./strings/io.jl:66
 [11] stringmime(::MIME{Symbol("image/svg+xml")}, ::Gadfly.Plot) at ./multimedia.jl:83
 [12] runner(::Type{Documenter.Expanders.ExampleBlocks}, ::Base.Markdown.Code, ::Documenter.Documents.Page, ::Documenter.Documents.Document) at /Users/tamasnagy/.julia/v0.6/Documenter/src/Expanders.jl:477
 [13] dispatch(::Type{Documenter.Expanders.ExpanderPipeline}, ::Base.Markdown.Code, ::Vararg{Any,N} where N) at /Users/tamasnagy/.julia/v0.6/Documenter/src/Selectors.jl:168
 [14] expand(::Documenter.Documents.Document) at /Users/tamasnagy/.julia/v0.6/Documenter/src/Expanders.jl:30
 [15] runner(::Type{Documenter.Builder.ExpandTemplates}, ::Documenter.Documents.Document) at /Users/tamasnagy/.julia/v0.6/Documenter/src/Builder.jl:178
 [16] dispatch(::Type{Documenter.Builder.DocumentPipeline}, ::Documenter.Documents.Document, ::Vararg{Documenter.Documents.Document,N} where N) at /Users/tamasnagy/.julia/v0.6/Documenter/src/Selectors.jl:168
 [17] cd(::Documenter.##2#3{Documenter.Documents.Document}, ::String) at ./file.jl:70
 [18] #makedocs#1(::Bool, ::Array{Any,1}, ::Function) at /Users/tamasnagy/.julia/v0.6/Documenter/src/Documenter.jl:203
 [19] (::Documenter.#kw##makedocs)(::Array{Any,1}, ::Documenter.#makedocs) at ./<missing>:0
 [20] include_from_node1(::String) at ./loading.jl:576
 [21] include(::String) at ./sysimg.jl:14
 [22] process_options(::Base.JLOptions) at ./client.jl:305
 [23] _start() at ./client.jl:371
while loading /Users/tamasnagy/.julia/v0.6/Gadfly/docs/make.jl, in expression starting on line 5
```

bjarthur:
not sure.  maybe try again now that we've merged a bunch of PRs?  thanks.

bjarthur:
@tlnagy did this error cause the entire build to fail, or just one example?  if the latter, just let it finish and then take a look at the as-built docs.  i'd like to proceed with finishing this PR if you like the direction it's headed.

Mattriks:
I looked at this (it built for me). In the gallery, I would like each code block do more than one plot (with `hstack`).  E.g. for each `Geom` I would like 2-3 single plots to be side by side (if a Geom has >3 plots, than using 2 code blocks = 2 rows of plots is ok). I think the plot size is right, but can the font size of the code blocks be reduced?, and hide the `using Gadfly, RDatasets` (but add at the top of the Gallery that the Gallery uses `RDatasets`). Also, the `set_default_plot_size` line is visible for me (I can see the `# hide` statement).  With these changes, the Gallery should look awesome!

bjarthur:
great idea about combing plots into subplots.  i've made the changed and pushed.

i'm not keen though on separating out the common code elements and putting them at the top.  i really want users to be able to make a single cut & paste to reproduce.

docs/src/gallery/geometries.md
```diff
@@ -0,0 +1,81 @@
...
+```@example
+using Gadfly, RDatasets, DataFrames
+set_default_plot_size(21cm, 8cm) # hide
+
+D = by(dataset("datasets","HairEyeColor"), [:Eye,:Sex], d->sum(d[:Freq]))
+p1 = plot(D, color="Eye", y="x1", x="Sex", Geom.bar(position=:dodge),
+          Guide.ylabel("Freq"));
+
+rename!(D, :x1 => :Frequency)
+palette = ["blue","brown","green","tan"]  # Is there a hazel color?
```
tlnagy: Leftover comment? This is unnecessary

tlnagy:
> did this error cause the entire build to fail, or just one example?

The whole build, but I pulled the latest changes and now I can't reproduce and everything is building 👍

> i'm not keen though on separating out the common code elements and putting them at the top. i really want users to be able to make a single cut & paste to reproduce.

I agree with @bjarthur here. 

> would be great if folks could checkout this PR, do a `julia docs/make.jl`, then open `docs/build/index.html` and have a look around.

Overall, I really like this! I think one thing that's missing is code examples in the docstrings for the Geoms/Guides/etc (e.g.  https://juliadocs.github.io/Documenter.jl/latest/lib/public/#Documenter.generate) so a user can see a concrete example of how a function can be used after running `?function`.

docs/src/gallery/geometries.md
Mattriks: Not leftover, it's a humorous comment!  Hazel is a color (think eyes/hazelnut),  and the plot is about eye color, so one should be able to use "hazel"!

Mattriks:
Great!  I particularly like the 3 plots for `Guide.xlabel` and `Guide.ylabel`. For the eye color plot in `Geom.bar`    I think `hstack(p1, p2a, p2b)` looks better than the double `hstack`. The palette needs to be changed to `[brown, blue, tan, green]` so that it matches the eye color labels. 

I would also like smaller font size for the code blocks, and/or where possible one-line plot statements e.g. for `Geom.point` p4 the `Geom.point` could be dropped from the plot statement, so it fits on one line.

docs/src/gallery/geometries.md
tlnagy: Ah. I get it.

bjarthur:
still working on this, and have come across a few things for which i'll submit issues for later.  just as a reminder for myself:

1. `Stat.violin` is very similar to `Stat.density`.  `Geom.violin` might be able to be refactored to us it instead.

2. what's the difference between `Geom.histogram(density=true)` and `Geom.density()`?  should we deprecate the former?

3. `order` in `Geom.line`, `Geom.polygon`, and `Geom.violin` would more descriptively be called `zorder`.  is it even necessary though given that z-order is set by the input argument order in `plot`?

4. why would one ever want to set `preserve_order` to `false` in `Geom.contour`?

5. should think about refactoring `struct RectangularBinGeometry` to use a default statistic of `Identity` and renaming it `RectangularGeometry`.  `Geom.rectbin` would then just use it as is, and `Geom.rectbin` would use `Stat.rectbin`.  as it stands now, `Geom.rect` overrides the default `Stat.rectbin` with `Stat.Identity`, which is backwards from all other geometries.  should also consider renaming `rectbin` to `tile` is [discussed](https://github.com/GiovineItalia/Gadfly.jl/pull/993#issuecomment-298177626) previously.

6. `Stat.binmean` doesn't appear to be used internally.  remove it?

tlnagy:
1. Agreed. I think it should be possible to get `Stat.density` to handle both cases. Also, if we make it generalizable enough then it should be possible to do something like https://github.com/clauswilke/ggridges too

2. `Geom.histogram(density=true)` just normalizes the histogram such that the integrated area is equal 1 so you can plot it on the same plot as a KDE. The name isn't intuitive though, something like `normalize` would make more sense, but density is standard across [matplotlib](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.hist.html) and ggplot2.

```julia
using Distributions

srand(123)
xs = rand(MixtureModel([Normal(0, 0.5), Normal(3, 1.2)]), 1000)

p1 = plot(layer(x=xs, Geom.density, Theme(default_color=colorant"red")),
     layer(x=xs, Geom.histogram), Guide.title("Geom.histogram(density=false)"))
p2 = plot(layer(x=xs, Geom.density, Theme(default_color=colorant"red")),
     layer(x=xs, Geom.histogram(density=true)), Guide.title("Geom.histogram(density=true)"))
draw(SVG(20cm, 10cm), hstack(p1, p2))
```

image

3. Shouldn't we just rename the `order` in Layer to `zorder` to be more explicit? And deprecate `order` in `line`, `polygon`, and `violin` because that `order` seems superseded by the `order` in Layer.

4. Not sure, I haven't used `Geom.contour`. Didn't @Mattriks write it? Maybe he knows.

5. That seems like a good plan.

6. I don't think we should remove, but it definitely needs some love. It wasn't even clear to be what is going on in the example plot here: http://gadflyjl.org/latest/lib/stats/stat_binmean.html#Stat.binmean-1. We could generalize `Stat.binmean` to make it like ggplot2's `stat_summary` geometry:  https://stackoverflow.com/questions/33150915/ggplot-plotting-the-bins-on-x-axis-and-the-average-on-y-axis. Currently `Stat.binmean` is specific case of `stat_summary`. I would fine with removing it too.

Mattriks:
3. Since some layers/Geoms use `z` e.g. `layer(z= , Geom.contour)`, I find `zorder` a bit confusing.
Also, the `order` in `Geom.line` (and other Geoms) is useful for the following reason.  
Each layer can have more than one Geom (reducing user code), so Geoms can be ordered separately *within* layers e.g. `Geom._(order=)` not `layer(order=)` is being used internally here:
```Julia
x = 0.0:0.1:2.0
D = DataFrame(x=x, ytrue=x.^2, y=(x.^2)+0.3*randn(length(x))  )
p = plot(D, x=:x, y=:y, ymin=D[:ytrue]-0.5, ymax=D[:ytrue]+0.5,
        layer(Geom.point, Geom.smooth(smoothing=0.9), Geom.ribbon)
        )  
```

4. I didn't write `Geom.contour`, and can't think of a reason to set `Geom.contour(preserve_order=false)`

5. Some big picture thinking might be useful here e.g. A `PolygonGeometry` is type of Geometry that allows overlapping polygons, whereas a `BinGeometry` bins data and renders as non-overlapping polygons.  `RectangularBinGeometry` currently mixes these 2 ideas. 

6. `Stat.binmean` is type of smoother, and would be of interest to smoothing & time series aficionados. Continuing from my example in point 3: 
```Julia
p = plot(D, x=:x, y=:y,  Geom.point,
        layer( x=:x, y=:y, Geom.line, Stat.binmean(n=10))
        )
```

Mattriks:
- `Scale.xgroup` & `Scale.ygroup` refer to `Geom.subplot_grid` axes. See my first example in #1125 
- `Scale.shape_discrete` see [here](https://github.com/GiovineItalia/Gadfly.jl/issues/766#issuecomment-298274527)

I don't have an example of `Scale.size_discrete`, but you can think of `Scale.(shape,size)_discrete` as being precursors to implementing `Guide.shapekey` and `Guide.sizekey` (#1133)

Mattriks:
And here is an example of `Scale.group_discrete()`:
```julia
geyser = dataset("datasets","faithful")
# Note in the next line the 2 groups are [true, false]
geyser[:g] = geyser[:Eruptions].>3.0

function plotscalegroup(scaleg::Scale.DiscreteScale)
    coord = Coord.cartesian(ymin=35, ymax=100)
    plot(geyser, coord, x=:Eruptions, y=:Waiting,  shape=:g, group=:g,
        Geom.point, Geom.ellipse,
        scaleg)
end

scales = [Scale.group_discrete(levels=[true, false]), Scale.group_discrete(order=[2,1])]    
plots =  plotscalegroup.(scales)   
hstack(plots...)
```
key_group



The right plot is simply to demonstrate that I can change the `group` order, but it has no visible effect.

In the left plot I should be able to do `Scale.group_discrete(levels=[true])`, to draw an ellipse only around the `[true]` group, but that currently fails on master due to #1130  (and `Geom.ellipse` is also on master)

bjarthur:
just added a commit to this PR to add the example in #602

tlnagy:
Awesome. I checked this out locally and I think it looks great. LGTM.

I think you and @Mattriks should be added to current maintainers on the docs main page and we should move @shashi to important contributer: https://github.com/GiovineItalia/Gadfly.jl/blob/master/docs/src/index.md

README.md
```diff
@@ -23,14 +23,27 @@ If you use **Gadfly** in a publication please consider citing it: [![DOI][citati
...
+```julia
+using Gadfly
+plot(y=[1,2,3])
+```
+
+## Gallery
+
+<div align="center"> <img
+src="https://cdn.rawgit.com/GiovineItalia/Gadfly.jl/master/docs/src/assets/gallery.png"
+alt="Gadfly Gallery" width="1024"></img> </div>
```
tlnagy: This doesn't seem to work? It looks like it should once you merge this PR.

bjarthur:
i was going to suggest we just give credit to dcjones (keep the first sentence), and then give the DOI link (replacing the other two).  the DOI has a list of contributors, which i think must be algorithmically computed based on git-blame.  this way the list won't go out of date so long as zenodo's bot works.  this paragraph and the DOI contributor list are already redundant with [github](https://github.com/GiovineItalia/Gadfly.jl/graphs/contributors) as it is.

tlnagy:
I think that's a good idea. Maybe rewrite the section to say:

"Gadfly is predominantly the work of Daniel C. Jones who initiated the project and built out most of the infrastructure. It is now maintained by a community of volunteers. Please consider citing it [zenodo link] if you use it in your work."

bjarthur:
to add to the list above:

7. should we rename `ElementOrFunction` to `ElementOrFunctionOrTheme` given its [definition](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/Gadfly.jl#L96)?

8. it's not clear to me what `set_default_plot_format` does, if anything.

9. much like interactive SVGs having multiple sets of ticks, only one of which is displayed depending on the zoom level, multiple sets of `Geom.label`s would be nice to have too.  the further you zoom in, the more labels that won't be overlapped, for the same font size, and so they should be dynamically made visible.

10. why is `spy` named as such, instead of just adding a new method to `plot`?

11. why alias `color_discrete` [to](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/scale.jl#L433) `color_discrete_hue`, and `color_continuous` [to](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/scale.jl#L529) `color_continuous_gradient` ?

12. [`GroupingScale`](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/scale.jl#L641) does not appear to be used internally.  remove?

13. `{x,y,z}_func`, `{x,y}_distribution`, `{shape,size}_identity` are all [defined](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/scale.jl#L667) as functions.  should we change them to be constants like all other `Scales`?  user code would be two characters more succinct without the parens.

14. `Shape.star{1,2}` both input `scalar` but then never use it.  the other shape that inputs `scalar` is `utriangle`, where it defaults to 1.  perhaps remove all three?

15.  `Geom.boxplot` currently only supports vertical plots, much like `Geom.violin`.

16. should we export `{current,with,pop,push)_theme`?  currently one must qualify the namespace with `Gadfly`.

Mattriks:
Quick comments:  

8. `set_default_plot_format(:png)` can be used in the same way as `set_default_plot_size()` e.g. can we use it in the Gadfly docs? (eg #1112)
9. Great!
10. Matlab `spy`
11. In my [`Scale`](https://github.com/Mattriks/Introduction_to_Gadfly_tutorials/tree/master/src) notebook, I suggest tidying `Scale.color_discrete` and `Scale.color_discrete_manual` e.g. by doing dispatch on or adding keywords to `Scale.color_discrete`. Having 2 or more separate functions seems untidy to me.
12. I agree
13. No comment at this point
14. `Shape.dtriangle` uses `scalar=-1`. I'm guessing someone thought about doing `Shape.ustar1`, `ustar2`, `dstar1` and `dstar2` - but didn't implement it.

bjarthur:
i think this PR is done.  please checkout this branch, compile the docs, and let me know if you have any comments.

tlnagy:
I think I have one main issue remaining with this PR. It appears to drastically cut back on the number of examples that we have per geometry, scale, etc. The gallery layout is nice for quickly looking for what geometry, etc you should use, but then it would be nicer to be able to dig deeper and see more examples for a specific geometry. I find `ggplot2`'s in depth docs per-geometry to be very helpful: http://ggplot2.tidyverse.org/reference/geom_density.html

Would it be possible to move the examples from their current location in `docs/src/libs/geom/etc` to the corresponding function docstring?

docs/src/gallery/guides.md
```diff
@@ -0,0 +1,101 @@
+# Guides
+
+## Guide.annotation
```
tlnagy: These should be ideally be links, e.g. `[Guide.annotation](@ref)`

bjarthur:
i'm confused...  the number of examples should be the same.  all i did was take the english text in `docs/src/libs/geom` and put it into a docstring in `src/geom` so that one could use the REPL's help system (the question mark command).  what remains are the examples, which i simply moved into `docs/src/gallery` and reformatted a bit.

tlnagy:
> i'm confused... the number of examples should be the same.

My bad, Github was autocollapsing the geometries gallery so my searching for examples wasn't coming up. 

In general, I'm a big fan of moving all the text to the docstrings and I think we should go ahead and merge this PR. I do however have some hesitation with how everything is laid out in the big monolithic files. I think in a future PR we should explore transitioning the gallery pages to strictly a gallery (i.e. only images, no code). Then clicking on the image or the text could open a dedicated page that has a lot more detail on the function and lots more examples. I'm a big fan of how seaborn handles this: https://seaborn.pydata.org/examples/index.html but I would prefer to have the names visible without hovering.

docs/src/gallery/guides.md
bjarthur: i've added links from the gallery to the library in the most recent commit.  thanks for the feedback!

Mattriks:
Here are some suggestions. In the Gallery:
- `Geom.bar` (2nd code block) do `hstack(p1, p2a, p2b)` rather than `hstack(p1, hstack(p2a, p2b))`
- `Scale.color_continuous` (2nd code block) do
```
x = repeat(collect(1:10)-0.5, inner=[10])
y = repeat(collect(1:10)-0.5, outer=[10])
```
so the grid fills the whole plot.
- `Scale.color_discrete_manual` (1st example) for me, this plot contains no points, but has a color key - do you see the points?

Mattriks:
And in the Library:
- `Gadfly` page: the code examples for `Compose.hstack` and `vstack` start with 3 ticks but don't end with 3 ticks

Mattriks:
In Development:
- The `Regression Testing` page needs an update (#1154)

bjarthur:
@Mattriks  have changed all your suggestions.  thanks for the thorough review!

the problem with the plot with the missing points is that the id number for elements is randomly generated.  when including multiple SVGs in the same HTML document they can conflict (ie be the same).  the hack was to change the seed.  long term we should look into longer random numbers, or creating a name space if possible.

i've started on a new "Thumbnail" section in the gallery, modeled after seaborn, but will put that in a separate PR as i don't want to hold this up anymore.

will merge this PR once CI passes.

PR #1128, opened 2018-04-04
Generate plots for stacks and layers documentation

### LINKED ISSUE #1127: Layers and stacks docs should generate plots
This was an oversight in my original doc update https://github.com/GiovineItalia/Gadfly.jl/pull/879 but the layers and stacks page doesn't have any generated plots. 

### Screenshot:

image

tlnagy:
This fixes an oversight in #879. Closes #1127.

In general, running all the code in the docs is good because we can catch any outdated examples.

### DIFF
--- a/docs/src/man/layers.md
+++ b/docs/src/man/layers.md
@@ -11,7 +11,14 @@ and stacking.
 
 Draw multiple layers onto the same plot with
 
-```julia
+```@setup 1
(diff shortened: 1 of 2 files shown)


tlnagy:
This branch fails to build locally, but it's unrelated to this PR so I think I'm going to go ahead and merge.

Issue #1125, opened 2018-04-01
broken example in docs


bjarthur:
@andreasnoack this [example](https://github.com/GiovineItalia/Gadfly.jl/blob/master/docs/src/lib/geoms/geom_ellipse.md) in the docs is broken, and the error messages are related to your dataframes PR.  could you please take a look?  thanks.

bjarthur:
possibly related, [test/testscripts/labels.jl](https://travis-ci.org/GiovineItalia/Gadfly.jl/jobs/360991398) is throwing a MethodError regarding discretize_make_ia when run by travis but not locally.  any ideas?

Mattriks:
& this example works in Gadfly 0.6.5, but is broken on master. Note that the species labels are in the expected order, but the plots don't match the labels (the top row is the default and is correct):
```Julia
iris = dataset("datasets", "iris")
sp = unique(iris[:Species])
Dhl = DataFrame(yint=[3.0, 4.0, 2.5, 3.5, 2.5, 4.0], Species=repeat(sp, inner=[2]) )

function plotxg(scalexg::Scale.DiscreteScale)
    plot(iris, xgroup=:Species,
        Geom.subplot_grid(
            layer( x=:SepalLength, y=:SepalWidth, Geom.point),
            layer(Dhl, xgroup=:Species, yintercept=:yint, Geom.hline(color="red", style=:dot) ),
            ),
    scalexg, 
    Theme(plot_padding=[0mm])
)
end

scales =    [Scale.xgroup(), Scale.xgroup(levels=["virginica","setosa","versicolor"]), Scale.xgroup(order=[3,1,2])]
plots = plotxg.(scales)

vstack(plots...)
```
![subplot](https://user-images.githubusercontent.com/18226881/38192518-a82e25ce-36b0-11e8-82ad-1dfe2bd27f35.png)

andreasnoack:
Thanks for reporting these. I know what is going on in @bjarthur's two examples. I'll try to look into this tonight.

Mattriks:
Also, same issue with `Scale.ygroup()`

Mattriks:
Again, top row correct, but in the other 2 rows, the class labels get sorted correctly, but boxplots don't: 
```Julia
function plotsxd(scalexd::Scale.DiscreteScale)
    plot(mpg, x=:Class, y=:Hwy, Geom.boxplot, scalexd,
        Theme(plot_padding=[0mm]) )
end 

[...]

PR #1139, opened 2018-04-24
make Geom.label consider point size

bjarthur:
positions of the labels are now adjusted not by a fixed padding of 2mm, but rather by the point size, either in theme.point_size or the size aesthetic.

@tlnagy you seem to be a frequent user of Geom.label.  can you please take a look?


tlnagy:
I don't use Geom.label that much so no strong opinions here. I'm okay with this implementation.

Also, I think we should tag a new release now that #1090 and #1125 are merged?

PR #1146, opened 2018-05-15
Use Gadfly.open_file instead of run(open) in the test suite (fix #1143)

### LINKED ISSUE #1143: `run open` in `compare_examples.jl` should be `open_file`
Regarding the issue about `run open` in test suit [raised here](https://github.com/GiovineItalia/Gadfly.jl/pull/1142#issuecomment-385285067). `Gadfly.open_file` is nice, but Gadly is huge and so fast to load. Hence the question, do we want to do `using Gadfly` in test suit (`compare_examples.jl` specifically)? Or maybe we simply copy-paste the definition of `open_file` inside it?

ulysses4ever:
Fixes #1143

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -88,6 +88,7 @@ include("aesthetics.jl")
 include("mapping.jl")
 include("scale.jl")
 include("theme.jl")
+include("open_file.jl")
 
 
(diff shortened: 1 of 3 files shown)


bjarthur:
thanks!

PR #1151, opened 2018-05-24
force x axis to be categorical for violin plots, fixes #968

### LINKED ISSUE #968: Geom.violin selects incorrect range for continuous variables
z_violin

When I save using:
```
draw(PNG("z_violin.png", 5inch, 4inch, dpi=300), z_plot)
```

Mattriks:
### Contributor checklist:

### Proposed changes
Fixes #968.  Realized there was a simple way to do this 🙂  
This minor change shouldn't cause changes anywhere else.

### DIFF
--- a/src/statistics.jl
+++ b/src/statistics.jl
@@ -1515,6 +1515,8 @@ end
 ViolinStatistic() = ViolinStatistic(300)
 
 input_aesthetics(::ViolinStatistic) = [:x, :y, :width]
+ Gadfly.default_scales(stat::ViolinStatistic) = [Gadfly.Scale.x_discrete(), Gadfly.Scale.y_continuous()]
+
 
 const violin = ViolinStatistic
 


tlnagy:
That's what I was looking for! However, I was thinking of extending violin to allow for horizontal violin plots and I don't think this strategy would work then, right?

tlnagy:
Actually it looks pretty simple. Something like what we do for `Geom.bar`

```julia
default_scales(stat::BarStatistic) = stat.orientation == :vertical ?
        [Gadfly.Scale.y_continuous()] : [Gadfly.Scale.x_continuous()]
```

Mattriks:
That will work, but for `violin` make one scale discrete.  This PR is ready to go, it's won't affect the docs!

tlnagy:
I'm gonna go ahead and merge this, but one difference in behavior between this PR and originally is the ordering:

```julia
using Distributions, DataFrames, Gadfly
srand(123);
data = vcat(rand(Beta(2,5), 1000), 1-rand(Beta(2,5), 1000), 0.25+rand(Beta(2,5), 1000));
group = vcat(fill(1, 1000), fill(0.5, 1000), fill(2, 1000)); df = DataFrame(data=data, group=group); p = plot(df, x=:group, y=:data, Geom.violin)
```

This PR

image

Master

image

I would imagine that if someone provides a continuous variables to `Geom.violin` then they would expect preservation of order when we convert to discrete.

Mattriks:
Since the default scale for `x` is now `Scale.x_discrete`, can change to a continuous scale by doing e.g.
```julia
plot(df, x=:group , y=:data, Geom.violin, Scale.x_continuous(minvalue=0.0, maxvalue=2.5) )
```

bjarthur:
+1 for extending to horizontal violins.  if you ever proceed with this, might take the time to get rid of `Stat.violin` and use `Stat.density` while you're at it.

Issue #1130, opened 2018-04-05
Scale.xdiscrete levels


Mattriks:
According to the [docs](http://gadflyjl.org/stable/lib/scales/scale_x_discrete.html), the `levels` argument of `Scale.x-discrete` behaves such that "Order will be respected and anything in the data that's not represented in levels will be set to NA".
- This is what I get on `Gadfly 0.6.5`:
```Julia
mpg = dataset("ggplot2","mpg")
ux = sort(unique(mpg[:Class]))
pa = plot(mpg, x=:Class, y=:Hwy, Geom.boxplot, color=:Class, Scale.x_discrete(levels=ux[1:4]))
pb = plot(mpg, x=:Class, y=:Hwy, Geom.boxplot, Scale.x_discrete(levels=ux[1:4]))
vstack(pa,pb)
```
![boxps2](https://user-images.githubusercontent.com/18226881/38341980-d335df72-38be-11e8-96e3-473d0a4280fa.png)
so it works as the docs suggest, but with obvious issues in the top plot, and maybe users would prefer for the NA column to not appear?
 
- And on master (for both plots):
 ````
ArgumentError: values not in levels encountered
Stacktrace:
 [1]discretize_make_ia(::CategoricalArrays.CategoricalArray{String,1,UInt8,String,CategoricalArrays.CategoricalString{UInt8},Union{}}, ::Array{String,1}) 
```

andreasnoack:
Thanks for capturing this. We actually discussed this behavior that results in the error above and we weren't sure what the right solution was. The handling of missing values is often not easy to spot in the code and I wouldn't be surprised if there were other issues with missing values. It probably isn't too hard to get the old behavior back (if we want it) but, as mentioned previously, it would probably be great if the handling of missing values was more centralized if possible.

Mattriks:
With a dataset containing 7 categories, if I use `levels` to specify 4 categories e.g. `Scale.x_discrete(levels=ux[1:4])` (as above) then the minimum behaviour I would expect is a plot with 4 levels, which is currently not what happens on master. 

 The NA label is added (by Gadfly) to the 3 levels I didn't specify (its not part of the raw data), which then becomes the first boxplot in the plots above. I'm also undecided whether that should be part of the default  behaviour.  But as mentioned, the minimum behaviour should be a plot with the 4 levels I specified.

bjarthur:
[...]

Issue #781, opened 2016-01-05
Geom.density isn't normed to 1


kain88-de:
Working with `Geom.density` I noticed that it doesn't normalize the density correctly.
See [Tower Sampling](https://gist.github.com/kain88-de/6208b0082359ddce404f), the two lines should overlap.

tlnagy:
The jupyter notebook provided doesn't run as is (it refers to an undefined variable `sample`). Normalization should be working correctly on master.

```julia
using Gadfly, Distributions
srand(123)
xs = linspace(-2, 2, 30)
dist = Normal(0, 1)
plot(layer(x=xs, y=pdf.(dist, xs), Geom.point, Theme(default_color=colorant"red")), layer(x=rand(dist, 10000), Geom.density), Coord.cartesian(xmin=-3, xmax=3))
```

![image](https://user-images.githubusercontent.com/1661487/40893909-618ed250-675a-11e8-8d98-0e7c9d7d4a25.png)

tlnagy:
Please open a new issue if you run into any problems.

PR #1157, opened 2018-05-28
[WIP] Density geometry revamp

### LINKED ISSUE #1152: Density geometries revamp
As discussed in #1151, `Gadfly.jl`'s current implementation of violin plots (and to a lesser extent, density plots) is fairly limited as compared to our equivalent ecosystems for other languages (e.g. [seaborn's violins](https://seaborn.pydata.org/generated/seaborn.violinplot.html) and [ggplot's violins](http://ggplot2.tidyverse.org/reference/geom_violin.html#aesthetics)). 

### Current limitations

- [ ] Not possible to change orientation of violinplots. The primary grouping currently has to be on the x axis so violins cannot be horizontal.

- [ ] Major duplication of code between `Stat.density` and `Stat.violin`

    - As pointed out by @bjarthur in #1151, there's major duplication going on here. Also it appears as though `ggplot2` is able to build on the same code.

- [ ] Not possible to stack densities

    - Three main options for position, (1) `:identity`: what we do now (2) `:stack` stack densities on top of each other like we do in `Geom.bar` (3) `:fill` normalize such that for each position along the densities, the total always equals 1

- [ ] Not possible to adjust scaling of density plots, i.e. all Gadfly violins/density plots always have an area of 1. It would help to add the option of using unnormalized counts.

- [ ] Currently impossible to group violins according to a 2nd variable (e.g. group by `color` and `x`)

    - This would need to be a generalization of @Mattriks' solution in #1081. Essentially, there are four cases: (1) if no color aesthetic is mapped, then use the default_color (2) if there is a 1 to 1 mapping between the color aesthetic and the primary grouping (e.g. the `x` categorical variable) then use the solution in #1081, (3) if there isn't a 1 to 1 mapping and fewer colors than x groupings then also use solution #1081 (4) group by primary variable, then by color (allowing split violins). 

- [ ] Several inefficiencies with the code (we could do quite a bit of preallocation that we don't do related to the KDE since we know how many datapoints there will be in advance)

### Examples

| Case | Example solution |
|:--|:--|
| Split violin grouped by x, color. Uses raw counts, area ≠ 1  | image|
| Case (3) for violin plots from above  | image |
| Horizontal violins  |  image |
| Use raw counts | image |
| Stacked densities |  image|
| `Fill`ed densities | image|

tlnagy:
Fixes #1152

# Contributor checklist:

# Implemented changes

- This PR introduces a bunch of new features for `Geom.density`: custom kernels, custom scaling of densities, stacking, trimmed distributions, vertical and horizontal orientations
- `Geom.violin` also gets all these features plus the ability to create split violins and group by color as well
- `Geom.violin` and `Geom.density` now share the same `DensityStatistic`. There is still some ugliness here because I infer whether the user wants a `density` or `violin` plot based on what variables are initialized in the aesthetic. I'm not wedded to this approach.
- This PR also introduces a general way for grouping by color and by category. I intend to leverage it for `Geom.boxplot` and maybe even `Geom.bar`.


tlnagy:
We now have the ability to generate conditional density distributions! 

```julia
using RDatasets, Gadfly
df = dataset("ggplot2", "diamonds")
plot(df, x=:Carat, color=:Cut, Geom.density(scale=:count, position=:fill))
```

image

@Mattriks and @bjarthur do you all know how I can adjust the x and y extents from an `apply_statistics` function? I would like to trim off the excess space in the above plot somehow.

bjarthur:
https://github.com/GiovineItalia/Gadfly.jl/issues/781 is related

tlnagy:
There's a bug in `KernelDensity.jl` that will sometimes cause this PR to fail. It would probably be best if my fix (https://github.com/JuliaStats/KernelDensity.jl/pull/52) is merged and tagged prior to merging this PR.

src/statistics.jl
```diff
@@ -500,67 +500,220 @@ function apply_statistic(stat::Density2DStatistic,
...
+end
 
-"""
-    Stat.density[(; n=256, bandwidth=-Inf)]
+output_aesthetics(stat::DensityStatistic) = [:x, :y, :color]
 
-Estimate the density of `x` at `n` points, and put the result in `x` and `y`.
-Smoothing is controlled by `bandwidth`.  Used by [`Geom.density`](@ref Gadfly.Geom.density).
-"""
-const density = DensityStatistic
```
bjarthur: to be consistent with the rest of the API, i'd suggest keeping `const density = DensityStatistic` and moving the docstrings for `struct DensityStatistic` to the retained`Stat.density`.

src/statistics.jl
```diff
@@ -500,67 +500,220 @@ function apply_statistic(stat::Density2DStatistic,
     apply_statistic(ContourStatistic(levels=stat.levels), scales, coord, aes)
 end
 
-
+"""
+    A general statistic for density plots (e.g. KDE plots and violin plots).
+See [`Geom.density`](@ref Gadfly.Geom.density) or [`Geom.violin`](@ref
+Gadfly.Geom.violin) for more details.
+"""
```
bjarthur: in #1116 i purposely tried to teach new users the GoG way of thinking by making the Geom docstrings explicitly defer explanation of details to the corresponding Stat docstrings for "[derived](https://github.com/GiovineItalia/Gadfly.jl/blame/master/docs/src/lib/geometries.md#L20)" geometries, not the other way around as you've done here.  i feel this is important as it really is a different way of doing things that many find hard to grasp.  does that make sense?

also, it would be good to be explicit about which aesthetics this statistic transforms, again so that the flow of data into making a graph is transparent.  so in this case `x` is transformed into `x` and `y` and is grouped by `color`.  in #1116 i defined [`aes2str`](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/Gadfly.jl#L72), to which one can input the output of `{input,output}_aesthetics()`,  to help ensure that docs don't get out of sync with code.  might not be useful in this case because of the grouping aesthetic in `:color`.

docs/src/gallery/geometries.md
```diff
@@ -102,13 +103,27 @@ using Gadfly, RDatasets, Distributions
...
+set_default_plot_size(21cm, 8cm)
+data = dataset("ggplot2", "diamonds")
+p1 = plot(data, x=:Carat, color=:Cut, Geom.density(position=:stack), Guide.title("Loses marginal densities"))
+p2 = plot(data, x=:Carat, color=:Cut, Geom.density(position=:stack, scale=:count), Guide.title("Preserve marginal densities"))
+hstack(p1, p2)
+```
+
+```@example
+using Gadfly, RDatasets
+plot(dataset("ggplot2", "diamonds"), x=:Carat, color=:Cut, Geom.density(position=:fill), Guide.title("Conditional density estimate"), Coord.cartesian(ymax=1.0, xmax=5))
```
bjarthur: this long line will likely create a horizontal slider in the generated doc html.  a hard line break would be nice.

src/geom/density.jl
```diff
@@ -0,0 +1,191 @@
...
+                                        scale, position, orientation, false)
+    DensityGeometry(stat, order, tag)
+end
+
+DensityGeometry(stat; order=1, tag=empty_tag) = DensityGeometry(stat, order, tag)
+
+"""
+   Geom.density(; bandwidth, adjust, kernel, trim, scale, position, orientation, order)
+
+Draws a kernel density estimate. This is a cousin of [`Geom.histogram`](@ref)
```
bjarthur: it would be nice to document how Geom.density(Stat.identity) behaved.  in other words, what aesthetics does Geom.density directly use.

src/geom/density.jl
```diff
@@ -0,0 +1,191 @@
...
+[`Stat.DensityStatistic`](@ref Gadfly.Stat.DensityStatistic) for details on
+optional parameters that can control the `bandwidth`, `kernel`, etc used.
+
+External links
+
+* [Kernel Density Estimation on Wikipedia](https://en.wikipedia.org/wiki/Kernel_density_estimation)
+"""
+const density = DensityGeometry
+
+element_aesthetics(::DensityGeometry) = Symbol[]
```
bjarthur: element_aesthetics should contain `:x`, `:y`, and `:color`, no?

bjarthur:
looking forward to having this functionality!  in general it looks great.

my only big question is whether separate Geoms are necessary (density and violin) with their own typedefs and render functions, or whether Stat.density could be combined with Geom.line, and Geom.density and Geom.violin could simply be aliases.

sorry for all the hassle resolving the conflicts with #1116.

bjarthur:
also, does this break anything?  is it backwards compatible with the former Geom.density and Geom.violin?  if not, depwarns are warranted.  and we should think about whether to merge it before or after fixing https://github.com/GiovineItalia/Gadfly.jl/issues/1130

tlnagy:
Thanks for the review. I won't have time to fix this for a bit and I want to make sure we get it right before merging. 

> also, does this break anything? is it backwards compatible with the former Geom.density and Geom.violin?

AFAIK, there shouldn't be any regressions or changes to the API. There are however some changes to defaults that I think are better than the current ones. The primary change is that density plots by default now clip to the range of the data (this the same as `ggplot2`) since it doesn't make sense to predict outside that range. 

> my only big question is whether separate Geoms are necessary (density and violin) with their own typedefs and render functions

This might be achievable. The primary difficulty is that these two geoms are grouped in very different ways.

- Density plots are generally overlaid or stacked on top of each other (varying color), only `x` and `y` aesthetics are used.
- Violin plots are generally dodged (or split) according to group and color, `x`, `y`, and `width` aesthetics are used.

They are both essentially `Compose.polygons`, but the offsets and spacing are different enough that gets a little hairy. If anything `Geom.boxplot` and `Geom.violin` are probably the most likely to share code.

I think in general we should think about separating out the positioning code into structs because a lot of it is boilerplate. The same positioning code should work for bars, boxplots, violins, etc. This is what `ggplot2` does. See http://ggplot2.tidyverse.org/reference/index.html#section-layer-position-adjustment for details

src/geom/density.jl
tlnagy: If I don't leave this blank, they are filled with autogenerated values so it's impossible to give useful error messages using `Gadfly.assert_aesthetics_defined`. I wasn't sure how to get around this so I leave this blank and figure out errors later: https://github.com/GiovineItalia/Gadfly.jl/pull/1157/files#diff-9ec506bf78232ae17d082c22c2e66449R616

src/geom/density.jl
tlnagy: Good point, I'll add this.

src/geom/density.jl
```diff
@@ -0,0 +1,191 @@
...
+using RDatasets, Gadfly
+
+df = dataset("ggplot2", "diamonds")
+
+p = plot(df, x=:Cut, y=:Carat, color=:Cut, Geom.violin())
+draw(SVG("diamonds_violin1.svg", 10cm, 8cm), p) # hide
+nothing # hide
+```
+![](diamonds_violin1.svg)
+"""
```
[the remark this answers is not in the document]
tlnagy: I think this would be nice to change, but it depends on https://github.com/JuliaDocs/Documenter.jl/issues/736 anyway so I'll remove this for now.

src/geom/density.jl
```diff
@@ -0,0 +1,191 @@
...
+    stat = Gadfly.Stat.DensityStatistic(n, bandwidth, adjust, kernel, trim,
+                                        scale, :dodge, orientation, true)
+    ViolinGeometry(stat, order, tag)
+end
+
+"""
+    Geom.violin[(; bandwidth, adjust, kernel, trim, order)]
+
+Draws a violin plot which is a combination of [`Geom.density`](@ref) and
+[`Geom.boxplot`](@ref). This plot type is useful for comparing differences in
```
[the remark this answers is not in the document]
tlnagy: I meant it stylistically, not technically, but that could change (see https://github.com/GiovineItalia/Gadfly.jl/pull/1157#issuecomment-395259725)

docs/src/gallery/geometries.md
tlnagy: Good catch, I had already corrected this locally.

src/statistics.jl
tlnagy: Makes sense. I'll correct this.

src/statistics.jl
tlnagy: Will do

bjarthur:
do you want to merge this before or after we support 0.7?  there will likely be merge conflicts if we wait.

tlnagy:
I don't have time right now to revamp this PR so I'm okay with letting it get a bit stale and then finalize it a bit later. There are enough small edge cases that I need to handle that it's better to sit on this till it's ready.

wizofe:
Split violin plot is a very useful function and it seems it's still not part of `Gadfly`. Any chance of possibility to revamp this PR? Let me know if I could help somehow.

bjarthur:
agreed!  @tlnagy ??

PR #1159, opened 2018-06-01
Use Iterators.flatten instead of IterTools.chain

iamed2:
# Contributor checklist:

- [ ] ~~I've updated the documentation to reflect these changes~~
- [ ] ~~I've added an entry to `NEWS.md`~~
- [ ] ~~I've added and/or updated the unit tests~~
- [ ] ~~I've `squash`'ed or `fixup`'ed junk commits with git-rebase~~

Most of the above is not applicable (this should be a non-functional change). I've successfully run `Pkg.test("Gadfly")`, but are there separate regression tests I need to worry about?

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -532,9 +532,9 @@ function render_prepare(plot::Plot)
         map(s->(s, _theme(l, plot)), collect(stats))
     end
 
-    for element in chain([(s, plot.theme) for s in plot.statistics],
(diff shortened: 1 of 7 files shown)


tlnagy:
I ran the regression tests and no difference was noted. Thanks for making this PR!

iamed2:
You're welcome, thanks for merging! Please check out the related https://github.com/GiovineItalia/Compose.jl/pull/285 if you have a chance :)

Issue #1166, opened 2018-06-19
Precompilation fails on Julia 0.6.3 with Windows


evanfields:
(Note that on Slack @ararslan suggested this might be fixed in Julia 0.6.4.)

Running Julia 0.6.3 on Windows 10, recent commits of Gadfly master won't compile. The last working commit seems to be c9113545, and with anything fec7e or after, precompilation does not succeed. (This is true after rebuilding all packages on which Gadfly depends, nuking the .ji files, etc.)

```
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.6.3 (2018-05-28 20:20 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-w64-mingw32

julia> Pkg.status("Gadfly")
 - Gadfly                        0.7.0+             c9113545

julia> Pkg.build("Gadfly"); using Gadfly
INFO: Building SpecialFunctions
INFO: Building Rmath
INFO: Building CodecZlib
INFO: Precompiling module Gadfly.

julia> versioninfo()
Julia Version 0.6.3
Commit d55cadc350* (2018-05-28 20:20 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: AMD Ryzen 5 1600X Six-Core Processor
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Zen)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, generic)
  
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: https://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.6.3 (2018-05-28 20:20 UTC)
 _/ |\__'_|_|_|\__'_|  |  Official http://julialang.org/ release
|__/                   |  x86_64-w64-mingw32

julia> Pkg.status("Gadfly")
 - Gadfly                        0.7.0+             fec7e18e

[...]

PR #1172, opened 2018-06-27
improve warning when nonunique colors used with manual_color_key

### LINKED ISSUE #1170: manual_color_key with non-unique colours throws unhelpful error
Using manual_color_key with non-unique colours throws the following unhelpful error.

```
BoundsError: attempt to access 1-element Array{Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}},1} at index [1:2]
macro expansion at dynamic.jl:67 [inlined]
(::Atom.##99#104)(::Dict{String,Any}) at eval.jl:74
handlemsg(::Dict{String,Any}, ::Dict{String,Any}, ::Vararg{Dict{String,Any},N} where N) at comm.jl:163
(::Atom.##11#13)() at event.jl:73
```

It took me a while to establish that the cause of this error was effectively due to the following generated code.

```julia
Guide.manual_color_key(legend_title, ["h", "Smooth"], ["black", "black"]),
```

In this case, I think it would be useful if Gadfly contained a sanity check.

tlnagy:
fixes #1170

### DIFF
--- a/src/guide.jl
+++ b/src/guide.jl
@@ -181,7 +181,7 @@ const colorkey = ColorKey
 # A helper for render(::ColorKey) for rendering guides for discrete color
 # scales.
 function render_discrete_color_key(colors::Vector{C},
-                         labels::OrderedDict{Color, AbstractString},
+                         labels::OrderedDict{C, AbstractString},
                          aes_color_label,
                          title_ctx::Context,
                          title_width::Measure,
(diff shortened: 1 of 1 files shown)


tlnagy:
Looks like tests are segfaulting on Travis. Not sure why. Tests are also segfaulting on https://github.com/GiovineItalia/Gadfly.jl/blob/f922f4d137a84cc9ff38e3cbd3b01847ca5b0a7e/test/testscripts/issue1125.jl#L23 locally for me on both this branch and master.

bjarthur:
the travis errors are similar to https://github.com/GiovineItalia/Gadfly.jl/issues/1166.  i wonder whether it would happen with julia 0.6.2.

i can confirm the segfault with `issue1125.jl` locally.

src/guide.jl
```diff
@@ -492,11 +492,18 @@ end
...
-    labels::Vector{AbstractString}
-    colors::Vector{C}
+    labels::OrderedDict{C, AbstractString}
+end
+function ManualColorKey(title, labels, colors::Vector{C}) where {C<:Color}
+    labeldict = OrderedDict{C, AbstractString}()
+    for (c, l) in zip(colors, labels)
+        if c in keys(labeldict)
+            error("Colors should not appear more than once")
+        end
```
bjarthur: `haskey(labeldict,c) || error()` would be more succint.  otherwise, lgtm once CI passes.

bjarthur:
https://github.com/GiovineItalia/Gadfly.jl/issues/1123 might be related

bjarthur:
worth noting that `issue1125.jl` works for me when run all by itself using `include("test/testscripts/issue1125.jl")`.   the failure only occurs for `Pkg.test("Gadfly")`.  so there must be something about the sequence of events leading up to it.

bjarthur:
https://github.com/JuliaLang/julia/issues/27607 seems to have the same error, and [the fix](https://github.com/JuliaLang/julia/pull/27568) is scheduled to be [backported](https://github.com/JuliaLang/julia/pull/27621) to 0.6.4.

tlnagy:
It looks like tests on v0.6 are passing on master again so I've updated this PR and hopefully CI passes this time.

tlnagy:
Looks like they're passing now!

Issue #1161, opened 2018-06-04
Use DocStringExtensions in Gadfly


tlnagy:
[`DocStringExtensions.jl`](https://github.com/JuliaDocs/DocStringExtensions.jl) is a really nice self-contained package that allows us to have much more flexible and dynamic docstrings. I'm very wary of adding more dependencies to Gadfly, but this package would add no subdependencies (it only depends on `Compat` and `Base`). A couple examples of where we can leverage it:

1) By using the automatically generated function signatures in the docstrings (so they don't get out sync) with the actual functions

2) By unrolling the fields in a type (see https://github.com/GiovineItalia/Gadfly.jl/pull/1157/files#diff-9ec506bf78232ae17d082c22c2e66449R507). Currently, all those docs are hidden when you do
`?Gadfly.Stat.density`, but could be unrolled with `DocStringExtension.jl`

Thoughts?

bjarthur:
auto-gen'd sigs sounds great, but where besides #1157 are docs embedded in typedefs?

bjarthur:
i can see the value in DocStringExtensions. https://github.com/GiovineItalia/Gadfly.jl/pull/1173 proposes to use it.

tlnagy:
Since it seems like we're generally onboard with using it, I'll go ahead and close this

Issue #852, opened 2016-06-07
Theme documentation outdated


Vectornaut:
The `Theme` documentation on the Gadfly web site, which I assume is produced from docs/themes.md, refers to the field `highlight_color`, which no longer exists. It has been replaced by `discrete_highlight_color` and `continuous_highlight_color`.

PR #1173, opened 2018-07-08
update index, tutorial, and manual

bjarthur:
1. made the index more succinct.
2. the `plot` signatures in the tutorial were way out of date.
3. moved the Array API from the manual to the tutorial
4. added `subplot_grid` to the manual.
5. use more interesting examples throughout
6. remove the reference to D3 in the backends section (no longer exists, right?)
7. make docstrings for `Theme` out of the text in the themes section.

once done it should be straightforward to convert the existing docs into a jupyter notebook for a video with @xorjane

closes https://github.com/GiovineItalia/Gadfly.jl/issues/1161

### DIFF
--- a/docs/src/man/themes.md
+++ b/docs/src/man/themes.md
@@ -4,236 +4,93 @@ Author = "Daniel C. Jones, Shashi Gowda"
 
 # Themes
 
-Many parameters controlling the appearance of plots can be overridden by passing
-a `Theme` object to the `plot` function. Or setting the `Theme` as the current theme using `push_theme` (see also `pop_theme` and `with_theme` below).
+Many parameters controlling the appearance of plots can be overridden by
+passing a `Theme` object to the `plot` function, or setting the `Theme` as the
+current theme using `push_theme` or `with_theme`.
 
-The constructor for `Theme` takes zero or more named arguments each of which overrides the *default* value of the field.
+The constructor for `Theme` takes zero or more keyword arguments each of which
+overrides the default value of the corresponding field.  See [`Theme`](@ref) for
+a full list of keywords.
+
+```@example 1
+using Gadfly, RDatasets
+set_default_plot_size(14cm, 8cm)  # hide
+mammals = dataset("MASS", "mammals")
+plot(mammals, x=:Body, y=:Brain, label=:Mammal,
+     Geom.point, Geom.label, Scale.x_log10, Scale.y_log10,
+     Theme(discrete_highlight_color=x->"red", default_color="white"))
+```
+
 
 ## The Theme stack
 
 Gadfly maintains a stack of themes and applies theme values from the topmost theme in the stack. This can be useful when you want to set a theme for multiple plots and then switch back to a previous theme.
 
-`push_theme(t::Theme)` and `pop_theme()` will push and pop from this stack respectively. You can use `with_theme(f, t::Theme)` to set a theme as the current theme and call `f()`.
+`push_theme(t::Theme)` and `pop_theme()` will push and pop from this stack
+respectively. You can also use `with_theme(f, t::Theme)` to temporarily set a
+theme as the current theme and call function `f`, which can be defined
+elsewhere, anonymously, or as a do-block.
 
-## `style`
-
-You can use `style` to override the fields on top of the *current* theme at the top of the stack. `style(...)` returns a `Theme`. So it can be used with `push_theme` and `with_theme`.
-
-
-## Parameters
-
-These parameters can either be used with `Theme` or `style`
-
-  * `default_color`: When the color aesthetic is not bound, geometry uses this
-    color for drawing. (Color)
-  * `point_size`: Size of points in the point, boxplot, and beeswarm geometries.  (Measure)
-  * `point_size_min`: Minimum size of points in the point geometry.  (Measure)
-  * `point_size_max`: Maximum size of points in the point geometry.  (Measure)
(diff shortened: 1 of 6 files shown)


docs/src/man/themes.md
```diff
@@ -4,236 +4,94 @@ Author = "Daniel C. Jones, Shashi Gowda"
...
-  * `label_out_of_bounds_penalty`: Used by `Geom.label(position=:dynamic)`
-  * `label_placement_iterations`: Used by `Geom.label(position=:dynamic)`
-  * `label_visibility_flip_pr`: Used by `Geom.label(position=:dynamic)`
-  * `label_hidden_penalty`: Used by `Geom.label(position=:dynamic)`
-  * `label_padding`: Used by `Geom.label(position=:dynamic)`
-
-## Examples
+The constructor for `Theme` takes zero or more keyword arguments each of which
+overrides the default value of the corresponding field.  See the Theme entry in
+the [Library](@ref Gadfly) for a full list of keywords.
```
tlnagy: Maybe have a direct link to Theme's docstring here?

docs/src/tutorial.md
```diff
@@ -8,30 +8,43 @@ Gadfly is an implementation of a "grammar of graphics" style statistical
...
 ```julia
-plot(data::DataFrame, mapping::Dict, elements::Element...)
+plot(data::AbstractDataFrame, elements::Element...; mapping...)
 ```
 
-The first argument is the data to be plotted, the second is a dictionary
-mapping "aesthetics" to columns in the data frame, and this is followed by some
-number of elements, which are the nouns and verbs, so to speak, that form the
+The first argument is the data to be plotted, the keyword arguments at the end
+map "aesthetics" to columns in the data frame, and everything in between is some
```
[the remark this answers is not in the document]
tlnagy: I second this.

docs/src/tutorial.md
```diff
@@ -60,22 +78,26 @@ the semi-colon trailing `plot`.  This automatically renders the image to
...
 to capture the output argument in this case.
 
+```julia
+plot(iris, x=:SepalLength, y=:SepalWidth)
 ```
-plot(iris, x=:SepalLength, y=:SepalWidth, Geom.point)
-```
+
+Note that there is also no need to specify `Geom.point`, as that will be
+supplied automatically if no other geometries are given.
```
tlnagy: Should we be recommending this behavior? IMO it's better to be explicit about what we're plotting.

docs/src/tutorial.md
```diff
@@ -88,20 +110,49 @@ renders delightful figures. Adding other geometries produces layers, which may
...
+## Arrays
+
+If by chance your data are stored in Arrays instead of a DataFrame, fear not,
+identical plots can be created using an alternate `plot` signature:
+
+```julia
+plot(elements::Element...; mapping...)
+```
+
+Here, the keyword arguments directly supply the data to be ploted,
```
tlnagy: > ploted

Should be plotted.

> instead of using them to indicate which columns of a DataFrame to use.

Not sure if this adds anything

docs/src/tutorial.md
```diff
@@ -88,20 +110,49 @@ renders delightful figures. Adding other geometries produces layers, which may
...
+```@example 1
+SepalLength = iris[:SepalLength]
+SepalWidth = iris[:SepalWidth]
+plot(x=SepalLength, y=SepalWidth, Geom.point,
+     Guide.xlabel("SepalLength"), Guide.ylabel("SepalWidth"))
+nothing # hide
+```
+
+Note that with the Array interface, extra elements must be included to specify the
+axis labels, whereas with a DataFrame they default to the column names.
```
tlnagy: > Note that with the Array interface, extra elements must be included to specify the axis labels, whereas with a DataFrame they default to the column names.

What about:

Note that with the Array interface, the axis labels cannot be directly inferred so include explicit calls to `Guide.xlabel` etc to specify the axis names.

src/theme.jl
```diff
@@ -58,88 +58,127 @@ function get_stroke_vector(linestyle::Symbol)
   error("unsupported linestyle: ", linestyle)
 end
 
+using DocStringExtensions
+
+"""
+$(FIELDS)
+"""
```
tlnagy: :tada: 

This will make keeping the docs in sync much easier and avoid issues like https://github.com/GiovineItalia/Gadfly.jl/issues/852

src/theme.jl
```diff
@@ -58,88 +58,127 @@ function get_stroke_vector(linestyle::Symbol)
...
 @varset Theme begin
     # If the color aesthetic is not mapped to anything, this is the color that
     # is used.
```
tlnagy: I feel like the comments are unnecessary with the docstring update and we should remove them.

src/varset.jl
```diff
@@ -58,3 +66,18 @@ macro varset(name::Symbol, table)
...
+    struct $(name)
+      $(docvar)
+      $(var)::$(typ)
+      #$(var)::$(typ)
+    end
+  end
+end
+"doc struct"
+@foo(Foo, "doc var", x, Int)
+=#
```
tlnagy: Where is this macro used?

bjarthur:
thanks for the reviews!  will merge once CI passes.

docs/src/tutorial.md
bjarthur: if we ever removed Geom.point as a default, then we should make this change.  otherwise best to document it i think.

tlnagy: Perhaps we could say something along the lines of:

> In cases where no explicit geometry is provided, the plot will use `Geom.point` by default.

This way, it doesn't sound like we're recommending this behavior, but it does document that this is the default.

PR #1156, opened 2018-05-28
Guide.shapekey

Mattriks:
- [x] I've updated the documentation to reflect these changes
- [x] I've added an entry to `NEWS.md`
- [x] I've added and/or updated the unit tests
- [x] I've run the regression tests
- [x] I've `squash`'ed or `fixup`'ed junk commits with git-rebase
- [x] I've built the docs and confirmed these changes don't cause new errors

### `Guide-ageddon`
This is the first step in `Guide-ageddon` (and closes #1133). This PR implements: 
- `Guide.shapekey(title=, labels=, pos=)`
- which generates a colored ShapeKey if `shape` and `color` are mapped to the same variable name. 
- `Theme(key_swatch_color=)`, useful if `shape` and `color` are mapped to different variables.

```julia
srand(123)
theme1 = Theme(point_size=3mm)
coord1 = Coord.cartesian(xmin=0.0, xmax=6.0)
D = DataFrame(x=1:5, y=rand(5), V1=["A","A","B","B","D"], V2 = string.([1,2,2,3,3])  )

pa = plot(x=1:5, y=[0.77, 0.94, 0.67, 0.39, 0.31], shape=["A","A","B","B","D"], theme1, coord1,
    Guide.shapekey(title="Key",labels=["α","β","δ"]),
    Guide.title("Guide.shapekey") )

# if shape==color, and you change levels, there are possible ways to auto-update the levels in Scale.color_discrete
# but for now, you have to do it manually:
pb = plot(D, x=:x, y=:y, shape=:V1, color=:V1,
        Scale.shape_discrete(levels=["D","A","B"]),     
        Scale.color_discrete(levels=["D","A","B"]), 
        theme1, coord1, Guide.title("Shape==Color") )

pc = plot(D, x=:x, y=:y, shape=:V1, color=:V2,  coord1,
        Guide.colorkey(title="Color", pos=[0.75w,-0.25h]),
        Guide.shapekey(title="Shape "),
        Theme(point_size=3mm, key_swatch_color="slategrey"),
        Guide.title("Shape!=Color")
        )
hstack(pa,pb,pc)
```
shapekey

To do in this PR:  
 - [x] Investigate auto-updating `Scale.color_discrete`, if `shape==color`, and you change `Guide.shapekey(levels=)`  (as in plot b above)

Future PRs (in no particular order):

[...]

PR #1118, opened 2018-03-18
Theme_colormaps

Mattriks:
The issues behind this PR are explained [here](https://github.com/GiovineItalia/Gadfly.jl/pull/1106#issuecomment-373900976). The current implementation in `Gadfly` whereby `Theme()` replaces the whole color scale has some flaws. This PR gives `Theme` control of just the color map functions, so the color scales (`Scale.color_discrete` and  `Scale.color_continuous`) can be used as normal. This is easier for both users and developers.  E.g. the `order` of key colors, and transformed color scales (e.g. `Scale.color_log10`) now behave as expected: 

```Julia
using Colors, RDatasets
dpalette(n::Int) = LCHab.(65, 100, 15:(360/n):374)
cpalette = [ LCHab(30, 70, 266), LCHab(100, 0, 108), LCHab(30, 70, 12)]

mytheme = Theme(discrete_colormap= dpalette, continuous_colormap= Scale.lab_gradient(cpalette...) )

Gadfly.with_theme(mytheme) do

pa = plot(dataset("datasets","iris"),
    x=:PetalWidth, y=:SepalLength, color=:Species, Geom.point,
    Scale.color_discrete(order=[3,2,1]),
    Guide.colorkey(title="Iris", pos=[0.1, 9.3])
)

pb = plot(dataset("ggplot2","diamonds"), x=:Price, y=:Carat,
    Geom.histogram2d(xbincount=25, ybincount=25),
    Scale.color_log10,
    Scale.x_continuous(format=:plain)
)
   
hstack(pa,pb)
end   
```  
mytheme

Also I've updated the `:dark` theme, added a `:ggplot` theme, and updated the docs.

PR #1175, opened 2018-07-14
Fixes #1126

### LINKED ISSUE #1126: Add doc example of transparency with Geom.ribbon
According to the documentation, it should be possible to make `Geom.ribbon` colors transparent by adding something like `Theme(lowlight_opacity=0.3)`. That does not appear to work however. This would be a great feature to have though.

Mattriks:

### Proposed changes
This PR (closes #1126):
- deprecates the`lowlight_opacity` argument in `Theme`, and adds a `depwarn`. 
- in the docs, updates `Geom.ribbon` and shows how to use `Theme(lowlight_color=)` to add color transparency to `Geom.ribbon`.
- enables `Theme(lowlight_color=)` for `Geom.polygon`. This makes plot 2 below possible:
```
using Colors, DataFrames, Distributions, Gadfly
x = -4:0.1:4
Da = [DataFrame(x=x, ymax=pdf.(Normal(μ),x), ymin=0.0, u="μ=$μ") for μ in [-1,1]]
Db = [DataFrame(x=randn(200)+μ, u="μ=$μ") for μ in [-1,1]] 

p1 = plot(vcat(Da...), x=:x, y=:ymax, ymin=:ymin, ymax=:ymax, color=:u, 
    Geom.line, Geom.ribbon, Guide.ylabel("Density"),
    Theme(lowlight_color=c->RGBA{Float32}(c.r, c.g, c.b, 0.4)), 
    Guide.colorkey(title="", pos=[2.5,0.6]), Guide.title("Parametric PDF")
)
p2 = plot(vcat(Db...), x=:x, color=:u, 
    Stat.density(bandwidth=0.5), Geom.polygon(fill=true, preserve_order=true),
    Coord.cartesian(xmin=-4, xmax=4),
    Theme(lowlight_color=c->RGBA{Float32}(c.r, c.g, c.b, 0.4)),
    Guide.colorkey(title="", pos=[2.5,0.6]), Guide.title("Kernel PDF")
)
hstack(p1,p2)
```
iss1126b
I added this example in the docs under `Stat.density`.

### DIFF
--- a/src/theme.jl
+++ b/src/theme.jl
@@ -173,11 +173,9 @@ $(FIELDS)
     "Color used to outline plot geometry. This is a function that alters (e.g. darkens) the fill color of the geometry. (Function)"
     continuous_highlight_color,     Function,        default_continuous_highlight_color
 
-    "Color used to draw background geometry, such as `Geom.ribbon`. This is a function that alters the fill color of the geometry.  (Function)"
+    "Color used to draw background geometry, such as `Geom.ribbon` and `Geom.polygon`. This is a function that alters the fill color of the geometry.  (Function)"
     lowlight_color,        Function,        default_lowlight_color
 
-    "Opacity of background geometry such as [`Geom.ribbon`](@ref).  (Float64)"
-    lowlight_opacity,      Float64,         0.6
 
     "Color altering function used to draw the midline in boxplots. (Function)"
     middle_color,          Function,        default_middle_color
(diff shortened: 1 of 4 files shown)


bjarthur:
any idea why tests are failing on shapekey?

also, given the removal of a field from Theme, we should probably bump the minor version when this is merged.  any other breaking changes we want to include in such a release?  maybe https://github.com/GiovineItalia/Gadfly.jl/pull/1118 ? 

otherwise, looks great!

Mattriks:
The `Theme()` constructor produced by the `@varset` macro allows strings to be parsed to colors.  You can tell this by doing: 
```
th = Theme(key_swatch_color="slategrey")
dump(th.key_swatch_color)
``` 
For the outer constructor that I added the field type e.g. `Color`

bjarthur:
i think you just need to add `parse_colorant` to https://github.com/GiovineItalia/Gadfly.jl/pull/1156

Mattriks:
Still trying to understand how to do this deprecation in a simple way. In order for `Theme(default_color="red")` to work, the `Theme` constuctor (containing any deprecations) needs to also be able to parse a string to colorant etc.  Note that `style(default_color="red")` does not work for the same reason i.e. `style` does not do any parsing of its `kwargs` before passing them to `Theme`.

src/theme.jl
```diff
@@ -393,3 +391,82 @@ end
...
+    label_visibility_flip_pr::Float64=0.2,
+    label_padding::Measure=1mm,
+    key_max_columns::Int=4,
+    discrete_color_scale::Scale.DiscreteColorScale=Scale.color_discrete(),
+    continuous_color_scale::Scale.ContinuousColorScale=Scale.color_continuous()
+)
+    
+    isfinite(lowlight_opacity) && Base.depwarn("The keyword argument `lowlight_opacity` has been deprecated, and never worked anyway!", :Theme)
+    
+return Theme(default_color, point_size, point_size_min, point_size_max, point_shapes, line_width, line_style, panel_fill, panel_stroke, panel_opacity, background_color,
```
bjarthur: would `Theme(default_color==nothing ? nothing : parse_colorant(default_color), ....` work?  this is how it's done for the non-deprecated constructor https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/varset.jl#L47

Mattriks: But that's no different then making a new `Theme` constructor, in which case the varset constructor for `Theme` is obsolete.  If the `Theme` constructor is coded explicitly then yes it easy to do deprecations. Removing the varset `Theme` constructor and making a new constructor should really be done as  a separate PR.

bjarthur: i think all we need here is

```
function Theme(; default_color, point_size, ..., lowlight_opacity, ...)
    Base.depwarn(....)
    Theme(
        default_color==nothing ? nothing : parse_colorant(default_color),
        point_size,
        ...
        #skip lowlight_opacity,
        ...)
end
```

no need to specify the types, as that will be handled by the varset definition of the Theme object.

bjarthur:
next tag will need to be a minor not point release since this introduces a breaking change.

bjarthur:
for some reason this PR is causing a warning, and i don't understand why.

```
julia> using Gadfly
WARNING: Method definition Type(Array{Any, 1}, Type{Gadfly.Theme}) in module Gadfly overwritten.
WARNING: Method definition (::Type{Gadfly.Theme})() in module Gadfly at /Users/arthurb/.julia/v0.6/Gadfly/src/varset.jl:62 overwritten at /Users/arthurb/.julia/v0.6/Gadfly/src/theme.jl:473.
```

the Theme constructor with the depwarn in it should be equivalent to the code below, but the latter doesn't emit the same warning.

```
julia> struct Foo
       a
       end

julia> Foo(; a=1) = Foo(a)
Foo

julia> Foo(; a=1, b=2) = Foo(a)
Foo
```

tlnagy:
I'm pretty sure this has something to do with varset creating a constructor and then the explicit constructor in `theme.jl` overwrites it. Might be better to do the deprecation in `@varset`, i.e. add an option to the macro to deprecate a field.

bjarthur:
think i've got a fix https://github.com/GiovineItalia/Gadfly.jl/pull/1180

Issue #1174, opened 2018-07-11
Add vertical bands like vspan() in Plots.jl


miguelmorin:
I would like to easily plot vertical bands, e.g. as a background for a recession in a time-series plot. `Plots.jl` has `vspan()`. In Gadfly, it requires a lot of code and mixes the color of the bands with the color of the time-series line.

For reference, the code is below, copied from [StackOverflow](https://stackoverflow.com/questions/51059491/how-to-easily-plot-background-vertical-bands):

```
using Gadfly, DataFrames, Colors

df = DataFrame(year = [1990; 2000], x = [0; 1], color = [1; 1])

x_shade = [1995 1995 1996 1996]
y_shade = [0 1 1 0]

theme = Theme(
    discrete_highlight_color = u -> ARGB(1, 1, 1, 0),
    default_color = colorant"grey")

p = plot(
    layer(
        x = x_shade,
        y = y_shade,
        Geom.polygon(preserve_order = true, fill = true),
        order = 1
    ),
    layer(df,
          y = "x",
          x = "year",
          color = "color",
          Geom.line,
          order = 2
          ),
    theme
)
```

bjarthur:
this produces identical output:

```
plot(df, y = "x", x = "year", color = "color", Geom.line,
     xintercept=[1995.5], Geom.vline(color="grey",size=[10]))
```

the only problem is that `size` can not currently be specified in the plot coordinate system.

miguelmorin:
In my specific case, that it is a problem: I am judging the behavior of a time-series depending around the edges of those boxes. So I would need to specify it the size in the original plot, e.g. `size = 1`. I would rather avoid hacks to convert between the plot coordinate system and the aesthetics system.

Mattriks:
`Geom.rect` can also be used:
```julia
using RDatasets, Gadfly

Dp = dataset("ggplot2","presidential")[3:end,:]
De = dataset("ggplot2","economics")
De[:Unemploy] /= 10^3
Dp[:y1] = 0; Dp[:y2] = 12
theme(x) = style(default_color=parse(Colors.Colorant, x))

p = plot(Dp, xmin=:Start, xmax=:End, ymin=:y1, ymax=:y2, Geom.rect, color=:Party,
    layer(De, x=:Date, y=:Unemploy, Geom.line, order=2, theme("black")),
    Scale.color_discrete_manual("deepskyblue", "lightcoral"),
[...]

PR #1176, opened 2018-07-21
HBandGeometry and VBandGeometry

CiaranOMara:
This feature provides hband and vband geometries that address a need similar to that expressed in #1174.

# Contributor checklist:

# Proposed changes
- Add hband geometry that has the same width properties of hline, but a vertical span specified by ymin and ymax.
- Add vband geometry that has the same height properties of vline, but a horizontal span specified by xmin and xmax.


bjarthur:
this provides a nice alternate parameterization to {h,v}line, where instead of specifying the intercept and width, you specify the min and max.  great to see unit tests.  still needs docstrings and a gallery addition.

@miguelmorin would you find this useful.

@tlnagy @Mattriks thoughts?

thanks for contributing @CiaranOMara .  and sorry for the delay getting back to you.  it's been a busy summer.

tlnagy:
Seems like a useful generalization of `Geom.vline` and `Geom.hline`. I was going to suggest that we make those a specific instance of these new geometries, but then we have to handle `Geom.abline` separately. I think it makes sense to organize it this way.

miguelmorin:
Thanks @CiaranOMara, and yes @bjarthur I would find this very useful.

CiaranOMara:
Thank you all for the feedback. 

I'll go ahead and develop the docstrings and gallery additions for this pull request. I'll then have a go at making `Geom.hline` and `Geom.vline` specific instances of these new geometries as well as separating `Geom.abline`.

I'm pretty busy this winter, but I hope to make another pull-request in the next 2 weeks.

Mattriks:
This could also perhaps be done as a `RectangleBinGeometry`, with `Geom.band` = `Geom.rect` + `Stat.band(orientation=)` .  That means the `color=:variable` aesthetic would work, as in my https://github.com/GiovineItalia/Gadfly.jl/issues/1174#issuecomment-404276855. Probably would need to make `Geom.rect` the base `RectangleGeometry`.

Mattriks:
Statistics work like this (this code is bare bones, just to illustrate the grammar of graphics). `Geo.rect` and `Geo.band`  have different `default_statistics`, and use the same `render` function,
```julia
module Geo
using Colors, Compose, DataFrames, Gadfly

struct RectangleGeometry <: Gadfly.GeometryElement 
    default_statistic::Gadfly.StatisticElement
    order::Int
    tag::Symbol 
end 

RectangleGeometry(default_statistic=Gadfly.Stat.identity(); order=0, tag=Gadfly.Geom.empty_tag) =
        RectangleGeometry(default_statistic, order, tag)

rect = RectangleGeometry

Gadfly.Geom.element_aesthetics(::RectangleGeometry) = [:xmin, :ymin, :xmax, :ymax, :color] 
Gadfly.Geom.default_statistic(geom::RectangleGeometry) = geom.default_statistic

band(; intercept=0.0, orientation=:vertical) =
    RectangleGeometry(BandStatistic(intercept, orientation))


function Gadfly.Geom.render(geom::RectangleGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics) 

    default_aes = Gadfly.Aesthetics() 
    default_aes.color = Gadfly.discretize_make_ia(RGBA{Float32}[theme.default_color]) 
    aes = Gadfly.inherit(aes, default_aes) 
    heights = aes.ymax-aes.ymin
    all(aes.ymax .== Inf) && (heights = fill(1h, length(aes.xmax)))
    
    ctx = context()
    compose!(ctx, rectangle(aes.xmin, aes.ymin, 
      aes.xmax-aes.xmin, heights, geom.tag), fill(aes.color)) 
    return compose!(context(order=geom.order), svgclass("geometry"), ctx)
end



struct BandStatistic <: Gadfly.StatisticElement
    intercept
    orientation::Symbol # :horizontal or :vertical like BarStatistic or HairStatistic
end
BandStatistic(;intercept=0.0, orientation=:vertical) = BandStatistic(intercept, orientation)

Gadfly.Stat.input_aesthetics(stat::BandStatistic) = [:xmin, :xmax, :ymin, :ymax]
Gadfly.Stat.output_aesthetics(stat::BandStatistic) = [:xmin, :xmax, :ymin, :ymax]
Gadfly.Stat.default_scales(stat::BandStatistic) = [Gadfly.Scale.x_continuous(), Gadfly.Scale.y_continuous()]

function Gadfly.Stat.apply_statistic(stat::BandStatistic,
                         scales::Dict{Symbol, Gadfly.ScaleElement},
                         coord::Gadfly.CoordinateElement,
                         aes::Gadfly.Aesthetics)
    if stat.orientation == :vertical
        aes.ymin = fill(stat.intercept, length(aes.xmin))
        aes.ymax = fill(Inf, length(aes.xmax))
    else
        aes.xmin = fill(stat.intercept, length(aes.ymin))
        aes.xmax = fill(maximum(aes.x), length(aes.ymax))
    end
end

end



using RDatatsets
Dp = dataset("ggplot2","presidential")
p = plot(Dp, xmin=:Start, xmax=:End, Geo.band(), color=:Party)

```
banda
There are obvious items to improve here e..g the handling of `ymin` and `xmin`.

CiaranOMara:
Thank you @Mattriks for the advice and examples. It was not immediately apparent to me that I needed to look in statistics.jl.

I can see that in general `Geom.rect` can be the base geometry for `RectangleBinGeometry` and perhaps any other rectangular element. But, I'm unsure where to reintroduce theme modifications, for example, the subtraction of `theme.bar_spacing`. You noticed in my earlier attempts that I had conceived of using `render` in a cascade.

In my implementation of BandStatistic, I decided to allow types of Measure to pass through (see [7eb4f](
https://github.com/GiovineItalia/Gadfly.jl/pull/1176/commits/7eb4f9fa9cefcbb361eb23f73bb57d877a068681)), but I'm not sure this is an agreeable solution. It assumes that the Measure is and will be correct for the render function.

bjarthur:
+100 for a solution which embraces GoG and uses a statistic.

why does Theme need to be modified?

also, @Mattriks would be great to having something like your post above on how statistics work in the development section of the manual

Mattriks:
Also great to see `Geom.rect` separated from the `BinGeometrys` (`hexbin` and `rectbin`). I also have a pca example in my `Statistics` notebook ([section 3.0](https://github.com/Mattriks/Introduction_to_Gadfly_tutorials/blob/master/src/Statistics.ipynb)) which I would like to put into in the Gadfly docs, to illustrate GoG concepts.

`Theme` doesn't need to be modified. The above post is simply saying that the new `render` function for `RectangularGeometry` doesn't use `theme.bar_spacing` as does `render(geom::RectangularBinGeometry)`. I'm ok with that because I wouldn't want to use `bar_spacing` with `Geom.band` - but what do others think?

@CiaranOMara have you done the [regression testing](http://gadflyjl.org/latest/dev/regression.html)? That is the first test whether your new coding has any side effects!

tlnagy:
> I'm ok with that because I wouldn't want to use `bar_spacing` with `Geom.band` - but what do others think?

Agreed. I think eventually we should have subsections inside of Theme for in-graph positioning, colors, fonts, etc because it would be nice to balance specificity with verbosity of `Theme`.

CiaranOMara:
I had the chance to run the regression tests to compare this development against the point at which this development forked. There are no differences between the SVGs and PNGs, except where I moved a call to `Geom.rect` from `testscripts/rectbin.jl` to `testscripts/rect.jl` (see https://github.com/GiovineItalia/Gadfly.jl/pull/1176/commits/a2648b12b531c8d215c389e30e291668af05c974#diff-7ccc6093f9e1518e18d85bfe70640ce0L10). I'm only able to report on PNGs and SVGs because I seem to have a rendering issue with PDFs and PSs at the fork and at the tip, which I hadn't noticed before. From what I can see, this development does not introduce any regressions/side effects. But, someone else might be in a better position to check PDFs and PSs.

CiaranOMara:
Building on an clarifying some of my of my points in relation to @tlnagy's comment.

> Agreed. I think eventually we should have subsections inside of Theme for in-graph positioning, colors, fonts, etc because it would be nice to balance specificity with verbosity of Theme.

The render function for `RectangularGeometry` repeats all of the code pertaining to colour in `RetangularBinGeometry`. I assume colour behaviour should be consistent for all rendered geometries. However, I haven't followed this back to see how other geometries handle colour. 

The other emergent constituent is the rendering of basic shapes; prominent in this example is the rectangle where the coordinates of its corners specify shape and position. All of the code that parses the coordinates and calls `Compose's Rectangle` can be the same. In the case of `RetangularBinGeometry`, the rectangle coordinates can be transformed as required for `theme.bar_spacing`, then passed on to some consistent code-block.

bjarthur:
this looks great so far.  a few suggestions:

1. delete `render(geom::RectangularBinGeometry...` from `src/geom/rectbin.jl` and have `Geom.rectbin` and `Geom.histogram2d` call the new `RectangularGeometry` in `src/geom/rect.jl`.  it's a near duplicate as you have said, the only difference being the two ` - theme.bar_spacing`.

2. workaround `Geom.{rectbin,histogram2d}` using `theme.bar_spacing` by adding another field to `RectangularGeometry`, which defaults to -0.05mm, and which `Geom.{h,v}band` overrides to be 0mm.  reference this field in the two places where `bar_spacing` was used.

3. remove `bar_spacing` from `Theme` and add a depwarn [like this](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/theme.jl#L473).

4. we should all consider renaming `Geom.rectbin` to `Geom.tile`, since it doesn't actually bin anything.  this would be more similar to ggplot's [nomenclature](https://ggplot2.tidyverse.org/reference/geom_tile.html).

src/statistics.jl
```diff
@@ -2037,4 +2038,49 @@ function Gadfly.Stat.apply_statistic(stat::EllipseStatistic,
...
+end
+
+input_aesthetics(stat::BandStatistic) = [:xmin, :xmax, :ymin, :ymax]
+output_aesthetics(stat::BandStatistic) = [:xmin, :xmax, :ymin, :ymax]
+default_scales(stat::BandStatistic) = [Scale.x_continuous(), Scale.y_continuous()]
+
+"""
+    Stat.band[(; orientation=:vertical)]
+
+Transform points in $(aes2str(input_aesthetics(band()))) into rectangles in
```
bjarthur: maybe "pairs of intercepts" instead of "points"?

test/testscripts/hvband.jl
```diff
@@ -0,0 +1,21 @@
+using Colors, Gadfly, RDatasets
+set_default_plot_size(21cm, 8cm)
+
+Dp = dataset("ggplot2","presidential")[3:end,:]
+De = dataset("ggplot2","economics")
+De[:Unemploy] /= 10^3
```
bjarthur: it's great to have nice examples like this that use real data sets in the documentation, but for tests i think it is better to have very simple test cases like in `testscripts/rect.jl`.  easier to debug if necessary, will execute more quickly, and create smaller files.

Mattriks:
Quick comments to the points above:
1-3. `Theme.bar_spacing` is also also used by `bar.jl`, so i don't agree that `bar.spacing` should be removed from `Theme`.
4. In terms of development, I think the BinGeometries (`hexbin` and `rectbin`) should be kept separate for the moment, so they can evolve separately from the PolygonGeometries (`rect` and `polygon`).

bjarthur:
in my mind, the BinGeoms are simply shortcuts for a Stat and a PolygonGeom.

and when you say rectbin, i think you mean histogram2d, no?  rectbin doesn't actually do any binning.

CiaranOMara:
I have split this pull request into two new pull requests so that discussions around `RectangularGeometry` can continue without hindering the addition of the new band geometries. 
I have placed `Geom.hband` and `Geom.vband` in #1264, and `RectangularGeometry` in #1263.

PR #1180, opened 2018-08-18
add way to deprecate Theme fields and fix 4542f3fb

bjarthur:
fixes https://github.com/GiovineItalia/Gadfly.jl/pull/1175#issuecomment-413979352 by adding another entry to the `varset` table that contains a `depwarn` string

### DIFF
--- a/src/varset.jl
+++ b/src/varset.jl
@@ -1,5 +1,8 @@
 # Generate large types where each field has a default value to which it's
-# initialized.
+# initialized.  the format of `table` is a symbol or tuple for each field, where
+# the former is just the variable name and the latter is of the form: ([doc str],
+# varable name, varable type, default value, [depwarn str]).  square brackets
+# indicate optional elements.
 
 macro varset(name::Symbol, table)
     @assert table.head == :block
@@ -7,6 +10,7 @@ macro varset(name::Symbol, table)
 
     names = Any[]
     vars = Any[]
+    depwarns = Any[]
     parsed_vars = Any[]
     parameters = Any[]
     parameters_expr = Expr(:parameters)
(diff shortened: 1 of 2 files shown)


src/varset.jl
```diff
@@ -7,39 +7,38 @@ macro varset(name::Symbol, table)
...
+            depwarnstr = ""
+        elseif row.head == :tuple && 2 <= length(row.args) <= 5
+            if isa(row.args[1], String)
+                docstr = row.args[1]
+                hasdocstr = true
+            end
+            var = row.args[hasdocstr+1]
+            typ = row.args[hasdocstr+2]
+            default = length(row.args)-hasdocstr > 2 ? row.args[hasdocstr+3] : :nothing
+            depwarnstr = length(row.args)-hasdocstr > 3 ? row.args[hasdocstr+4] : ""
```
tlnagy: So the only thing that determines whether something is deprecated or not is whether it has another docstring line following it? I was thinking that you could explicitly use syntax similar to the @deprecated macro (like explicitly have @deprecated in the first position of the line with the alt text) and then to do some processing to determine whether it's present. It would be nice to be super clear that something is deprecated when looking at the source.

src/varset.jl
```diff
@@ -59,7 +58,10 @@ macro varset(name::Symbol, table)
             $(vars...)
         end
 
-        $(name)($(parameters_expr)) = $(name)($(parsed_vars...))
+        function $(name)($(parameters_expr))
+            $(depwarns...)
+            return $(name)($(parsed_vars...))
+        end
```
tlnagy: I like this approach and this should scale well in the future.

src/varset.jl
bjarthur: i've added docs to the top of src/varset.jl to explain the format of the table.

PR #1183, opened 2018-08-19
simplify and correct varset docs

bjarthur:
as it stands now, to deprecate a field one must supply a type and default value.  good enough for now i think.

### DIFF
--- a/src/varset.jl
+++ b/src/varset.jl
@@ -1,8 +1,7 @@
-# Generate large types where each field has a default value to which it's
-# initialized.  the format of `table` is a symbol or tuple for each field, where
-# the former is just the variable name and the latter is of the form: ([doc str],
-# varable name, varable type, default value, [depwarn str]).  square brackets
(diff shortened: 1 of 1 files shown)


src/varset.jl
```diff
@@ -1,8 +1,7 @@
-# Generate large types where each field has a default value to which it's
-# initialized.  the format of `table` is a symbol or tuple for each field, where
-# the former is just the variable name and the latter is of the form: ([doc str],
-# varable name, varable type, default value, [depwarn str]).  square brackets
-# indicate optional elements.
+# Generate large types where each field has an optional default value
+# to which it's initialized.  the format of each row in `table` is
+# "[doc str], varable name, [varable type, [default value, [depwarn
+# str]]]", where square brackets indicate optional elements.
```
Mattriks: spelling of variable?

bjarthur: right!  thanks.

PR #1186, opened 2018-08-21
permit more than 256 discrete levels

bjarthur:
a partial solution to https://github.com/GiovineItalia/Gadfly.jl/issues/1130#issuecomment-380966426

i still need to run CI, regression, and add a unit test.

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -392,8 +392,7 @@ function trim_zip(xs...)
 end
 
 # Convenience constructors of IndirectArrays
-discretize_make_ia(values::AbstractVector, levels) =
-    IndirectArray(Array{UInt8}(indexin(values, levels)), levels)
+discretize_make_ia(values::AbstractVector, levels) = IndirectArray(indexin(values, levels), levels)
 discretize_make_ia(values::AbstractVector)         = discretize_make_ia(values, unique(values))
 discretize_make_ia(values::AbstractVector, ::Void) = discretize_make_ia(values)
 
--- a/test/testscripts/levels.jl
+++ b/test/testscripts/levels.jl
@@ -1,0 +1,5 @@
+using Gadfly
+
+n=16
+pts = repeat(1:n,inner=(1,n))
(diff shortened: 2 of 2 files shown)


src/misc.jl
```diff
@@ -392,8 +392,8 @@ function trim_zip(xs...)
 end
 
 # Convenience constructors of IndirectArrays
-discretize_make_ia(values::AbstractVector, levels) =
-    IndirectArray(Array{UInt8}(indexin(values, levels)), levels)
+discretize_make_ia(values::AbstractVector, levels::AbstractVector{T}) where T =
+    IndirectArray{T,1,AbstractVector{UInt32},AbstractVector{T}}(indexin(values, levels), levels)
```
tlnagy: Shouldn't using UInt work here and letting Julia pick the optimal bitstype?

bjarthur: UInt is an alias for UInt64

bjarthur:
@andreasnoack @nalimilan could you please review? 

CI and regression tests pass.  just need to add a test.

bjarthur:
and @Mattriks , does this PR close https://github.com/GiovineItalia/Gadfly.jl/issues/1130 ?

Mattriks:
I'm assuming this doesn't fix the first issue: https://github.com/GiovineItalia/Gadfly.jl/issues/1130#issue-311438805. That is, where you have a dataset with _p_ levels, but you use the `levels` arg to specify only _q_ levels  where _q_<_p_. Did you test with this PR?

bjarthur:
correct @Mattriks .  that's more of a design decision, and i'm just trying to fix the regression you found later.

nalimilan:
@bjarthur Have you noticed my comment?

Mattriks:
The issue for me is that the ability to specify fewer levels than available levels worked in Gadfly 0.6.5, but doesn't work on master. As this does not work on master than this statement in the docs:
"Order will be respected and anything in the data that's not represented in `levels` will be set to missing." needs to say :
"Order will be respected but attempting to set select fewer levels than available levels is currently not working."

PR #1189, opened 2018-08-26
support julia 0.7

bjarthur:
fixes https://github.com/GiovineItalia/Gadfly.jl/issues/1158, https://github.com/GiovineItalia/Gadfly.jl/issues/1185

PR #1193, opened 2018-09-14
Fix test failure in discretize_make_ia

nalimilan:
This PR is just to test a few possibles fixes/workarounds for the failure at https://github.com/GiovineItalia/Gadfly.jl/pull/1189.

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -348,7 +348,11 @@ function trim_zip(xs...)
 end
 
 # Convenience constructors of IndirectArrays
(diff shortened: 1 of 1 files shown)


bjarthur:
thanks for looking into this!

nalimilan:
Actually it turns out this was trivial. I have no idea why I haven't realized the missing `something` call before, I could have sweared it was present.

There's a problem when trying to build docs, maybe related to the use of `SubString`? Can be fixed later anyway.

PR #1207, opened 2018-10-10
Fix deprecated warn() calls, whitespaces and a Compat usage

laborg:
Just a couple of minor fixes. Found while using master on 1.0.1. 
All tests pass except one (for the TranscodingStream-RDatasets issue)

A review would be welcomed.

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -49,7 +49,7 @@ function __init__()
         try
             push_theme(Symbol(strip(theme)))
         catch err
-            warn("Error loading Gadfly theme $theme (set by GADFLY_THEME env variable)")
+            @warn "Error loading Gadfly theme $theme (set by GADFLY_THEME env variable)"
             show(err)
         end
     else
@@ -479,8 +479,8 @@ function render_prepare(plot::Plot)
 
     defined_unused_aesthetics = setdiff(mapped_aesthetics, used_aesthetics)
     isempty(defined_unused_aesthetics) ||
-            warn("The following aesthetics are mapped, but not used by any geometry:\n",
-                join([string(a) for a in defined_unused_aesthetics], ", "))
+            @warn "The following aesthetics are mapped, but not used by any geometry:\n" * 
+                join(defined_unused_aesthetics, ", ")
 
     scaled_aesthetics = Set{Symbol}()
     for scale in plot.scales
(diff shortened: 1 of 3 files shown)


src/Gadfly.jl
```diff
@@ -49,7 +49,7 @@ function __init__()
         try
             push_theme(Symbol(strip(theme)))
         catch err
-            warn("Error loading Gadfly theme $theme (set by GADFLY_THEME env variable)")
+            @warn "Error loading Gadfly theme $theme (set by GADFLY_THEME env variable)"
```
bjarthur: strange that femtocleaner didn't catch this

src/Gadfly.jl
```diff
@@ -479,8 +479,8 @@ function render_prepare(plot::Plot)
 
     defined_unused_aesthetics = setdiff(mapped_aesthetics, used_aesthetics)
     isempty(defined_unused_aesthetics) ||
-            warn("The following aesthetics are mapped, but not used by any geometry:\n",
-                join([string(a) for a in defined_unused_aesthetics], ", "))
+            @warn "The following aesthetics are mapped, but not used by any geometry:\n" * 
+                join(defined_unused_aesthetics, ", ")
```
bjarthur: nice that `join` autoconverts its arguments to strings!  new to me.

src/mapping.jl
```diff
@@ -61,7 +61,7 @@ function cleanmapping(mapping::Dict)
         if haskey(aesthetic_aliases, key)
             key = aesthetic_aliases[key]
         elseif !in(key, fieldnames(Aesthetics))
-            warn("$(string(key)) is not a recognized aesthetic. Ignoring.")
+            @warn "$(string(key)) is not a recognized aesthetic. Ignoring."
```
bjarthur: interpolations is, or it least was IIRC, slow.  so in cases like this i believe `string(key, " is not a recognized aesthetic. Ignoring.")` is preferred.

laborg: I don't think interpolation speed has changed, but in this case I prefer less noise over performance as the warning isn't actually called anywhere near a hot loop.  I now see that it would be better to have just write: `@warn "$key is not a recognized aesthetic. Ignoring."`

laborg: But if you prefer performance, I'll update the PR.

bjarthur:
thanks for this!

PR #1217, opened 2018-10-25
Update to Documenter 0.20

mortenpi:
Not sure if this will actually work.

Due to upcoming breaking changes in Documenter. This makes sure that the doc builds do not start failing when 0.20 lands in METADATA. [See this thread on Discourse for more information.](https://discourse.julialang.org/t/psa-documenter-jl-breaking-changes-version-capping/16431) 

Ref: JuliaDocs/Documenter.jl#861

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ jobs:
         - |
           julia --color=yes --project=docs/ <<EOF
             using Pkg
+            Pkg.add(PackageSpec(name="Compose", rev="master"))
             Pkg.develop(PackageSpec(path=pwd()))
             Pkg.instantiate()
             include("docs/make.jl")
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,18 @@ before_script:
   - julia -e 'using Pkg; Pkg.add(PackageSpec(name="Compose", rev="master"))'
 after_success:
   - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
-  - julia -e 'using Pkg; foreach(x->Pkg.add(x), readlines(open(joinpath("docs", "REQUIRE")))); include(joinpath("docs", "make.jl"))'
+
+jobs:
+  include:
+    - stage: "Documentation"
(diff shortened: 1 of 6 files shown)


mortenpi:
Ok, it didn't work. I might get back to this in a few days.

tlnagy:
Thanks for the PR, did you get a chance to look into this more?

mortenpi:
Apologies for not getting back to this earlier. So, instead of capping Documenter to 0.19, I've simply updated `make.jl` so that it would work with the latest version. I also switched to using `docs/Project.toml` to list the docs-only dependencies.

If Travis passes, it should be good to go. It is also using a build stage now, which means that documentation build failures should also fail Travis now by failing the "Documentation" build stage.

.travis.yml
```diff
@@ -14,4 +14,18 @@ before_script:
   - julia -e 'using Pkg; Pkg.add(PackageSpec(name="Compose", rev="master"))'
```
mortenpi: Hmm.. the dependency in `docs/Project.toml` overrides this. Is it important that Compose is on `master` for documentation building as well?

bjarthur: in general, yes.  sometimes changes to gadfly master depend on changes to compose master that haven't been tagged yet.

mortenpi: Alright. I did not see how I could do this without duplication, so I just added the same `Pkg.add` into the docs building command.

mortenpi:
I forgot about this before -- this also makes another change. Currently, the docs from `master` end up under the `http://gadflyjl.org/latest/` URL, but with this PR they start deploying to `http://gadflyjl.org/dev/`. This is the new recommended default but it can be changed.

If you are OK with changing the URL, I can update the links in the README right away. If not, I'll add a keyword to `deploydocs` to keep it at `latest/`.

The latest tagged version will still be available from `stable/`, so no change to that.

bjarthur:
i'm find with dev/ instead of latest/.  make the change to the REAME and i'll merge.

mortenpi:
Done.

bjarthur:
thanks!

PR #1224, opened 2018-11-14
Use -e option instead of heredoc in CI doc stage

mortenpi:
Follows up on #1217. I realized that the way I set up the Documentation build stage, it does not actually fail the stage if documentation building errors -- this fixes that. The problem is that Julia does not set the return code appropriately when the scripts are passed via stdin (e.g. as a heredoc, as was here; ref: JuliaLang/julia#30039).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,11 +22,11 @@ jobs:
       os: linux
       script:
         - |
-          julia --color=yes --project=docs/ <<EOF
+          julia --color=yes --project=docs/ -e'
             using Pkg
             Pkg.add(PackageSpec(name="Compose", rev="master"))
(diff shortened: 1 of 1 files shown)


bjarthur:
as it happens the doc CI is failing b/c it relies on a specific version of Compose instead of the latest master.  happy to merge once this is fixed.  thanks for all this!

mortenpi:
Sorry, I did not notice the CI failure (which is a bit ironic :laughing:). It did not appear to be Compose-related, rather a problem with Pkg since `Compose#master` was being installed twice. It should be fixed now, and the docs appear to build fine with Compose master.

PR #1229, opened 2018-12-17
Tweak mapping to improve plotting performance

non-Jedi:
This was a learning exercise for me to practice using Julia profiling tools. I picked Gadfly's plot function as it seemed likely to have some low hanging fruit while being relatively simple to grok. These changes improve the benchmark posted below on my system from 18.7 microseconds to 10.3 microseconds through minor refactoring; definitely not earth-shattering but in the direction of goodness. I leave it to the maintainers discretion whether it's worth their time to review this patch.

Note that this only improves plot generation and not plot display, which takes the bulk of the time in an interactive plotting session.

```julia

julia> using Gadfly, BenchmarkTools

julia> x = 1:10000; y = rand(1:100, 10000); c = rand(1:250, 10000);

julia> @btime plot(x=$x, y=$y, color=$c);
  10.281 μs (39 allocations: 4.44 KiB)
```

# Contributor checklist:

-
-
-

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -165,10 +165,9 @@ plot(ls..., Guide.title("layer example"))
 """
 function layer(data_source,
                elements::ElementOrFunction...; mapping...)
-    mapping = Dict{Symbol, Any}(mapping)
     lyr = Layer()
     lyr.data_source = data_source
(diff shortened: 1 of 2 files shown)


bjarthur:
there's a nobel prize for whoever figures out the time-to-first problem.  thanks for this!

PR #1232, opened 2018-12-30
support TerminalExtensions.jl

### LINKED ISSUE #1230: display()
old versions of Gadfly for julia 0.6,4 were producing plots that can be plotted by TerminalExtensions.jl in the REPL. The new versions for 0.7 & 1.0 cannot do this.

Function as this is needed:

```import Base: display
function display(p::Gadfly.Plot)
Gadfly.draw(Gadfly.PNG(), p)
println()
end
```

can you add something like this in the code?

bjarthur:
Fixes #1230 @montyvesselinov

ideally it would be nice to replace the `popdisplay(GadflyDisplay)` in src/terminalextensions.jl with code to put `GadflyDisplay` before `iTerm2.InlineDisplay`.  as a fallback in case you have TerminalExtensions in your startup.jl, but are using a terminal which is not supported.  one can get a `typeof(Base.Multimedia.displays[i])`, but how does one tell which module that type was defined in?

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -41,6 +41,11 @@ export Plot, Layer, Theme, Col, Row, Scale, Coord, Geom, Guide, Stat, Shape, ren
 export SVGJS, SVG, PGF, PNG, PS, PDF, draw, inch, mm, cm, px, pt, color, @colorant_str, vstack, hstack, title, gridstack
 
 
+function link_terminalextensions()
+    @debug "Loading TerminalExtensions support into Gadfly"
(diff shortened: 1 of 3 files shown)


bjarthur:
@montyvesselinov  can you please see if this PR works for you?  i just pushed some changes that use dispatch to put the iTerm2 display at the end of `Multimedia.displays`, instead of deleting the `GadflyDisplay` at the end, so that folks who have put TerminalExtensions.jl in their startup can still use Gadfly in another type of terminal without error.

PR #1239, opened 2019-01-06
Fixes #1218

### LINKED ISSUE #1218: subplot_grid not working with Scale
Hi, 
I am using Julia 1.0.1 with Gadfly v1.0.0

The following plot does not render with a Scale. 
`p1 = plot( filter(row -> row[:y] < 99, tmp_hb ), 
    ygroup =:c3, x=:x1, y=:y1,
    Scale.color_continuous(colormap=Scale.lab_gradient("blue", "white", "orange")),
    Geom.subplot_grid(Geom.hexbin)
)`
In Julia v 0.6, it would render with one scale for the entire plot (which is what I want).

`p1 = plot( filter(row -> row[:y] < 99, tmp_hb ), 
    ygroup =:c3, x=:x1, y=:y1,
    Geom.subplot_grid(Scale.color_continuous(colormap=Scale.lab_gradient("blue", "white", "orange")), Geom.hexbin)
)`
Produces one Scale per subplot, which is what I expected.

Is this a bug or am I missing something obvious. 
Thanks
Peter

Mattriks:

### This PR:

- Fixes #1218
- `Geom.subplot_grid(Geom.contour())` also doesn't behave in the same way as for Julia 0.6. This PR fixes this issue by instead using: `Geom.subplot_grid(layer(Stat.contour, Geom.line))`,
and throws a warning if a user tries `Geom.subplot_grid(Geom.contour())`
- Adds `Stat.contour` example in the plot gallery 

### Example #1218
```julia
using DataFrames, Distributions, StatsBase
N = 1000
df1 = DataFrame(
  x1 = rand(Normal(1,3), N),
  x2 = sample(["High", "Medium", "Low"], pweights([0.25,0.45,0.30]), N),
  x3 = rand(Normal(2, 6), N)
)

p = plot(df1, xgroup =:x2,  x=:x1,  y=:x3,
    Geom.subplot_grid(Geom.hexbin(xbincount=20, ybincount=20)),
    Guide.xlabel(nothing)
)
```
issue1218

### DIFF
--- a/docs/src/gallery/statistics.md
+++ b/docs/src/gallery/statistics.md
@@ -12,6 +12,23 @@ p2 = plot(dataset("datasets", "iris"), x="SepalLength", y="SepalWidth",
 hstack(p1,p2)
 ```
 
+## [`Stat.contour`](@ref)
+
+```@example
+using DataFrames, Gadfly
+set_default_plot_size(21cm, 8cm)
+expand_grid(xv, yv, n::Int) = vcat([[x y z] for x in xv, y in yv, z in 1:n]...)
+a = expand_grid(6.0:10, 1.0:4, 3)
+D= DataFrame(x= a[:,1], y=a[:,2], z=a[:,3].*a[:,1].*a[:,2], 
+    level=string.("L",floor.(Int, a[:,3])) )
+coord = Coord.cartesian(xmin=6, xmax=10, ymin=1, ymax=4)
+
+plot(D, xgroup=:level, x=:x, y=:y, color=:z,
+    Geom.subplot_grid(coord, layer(z=:z, Stat.contour(levels=7), Geom.line)),
+    Scale.color_continuous(minvalue=0, maxvalue=120)
+)
+```
+
 ## [`Stat.density`](@ref)
 
 ```@example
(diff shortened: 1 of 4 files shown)


bjarthur:
can `Geom.contour` really not be fixed in 0.7 to behave the same way as in 0.6?  seems like a hack (hopefully temporary until we fix it) to ask the user to use `Geom.line; Stat.contour` instead.

Mattriks:
I think these issues are due to scoping changes, and`Geom.subplot_grid` may need a rewrite (which will take a while). In the meantime this PR gives users the functionality.

bjarthur:
why is it that `Geom.subplot_grid(Geom.hexbin)` works but `Geom.subplot_grid(Geom.contour)` does not?  are there other geometries that don't work?

i'm hesitant to merge a temporary fix like this, as a more permanent solution might then never happen.

Mattriks:
Do you know a `Geom.subplot_grid(Geom.hexbin)` example is used in this new book?:
https://discourse.julialang.org/t/data-science-with-julia/19941
I hope that @Aeonglacial posts a selfie after trying the example in Section 4.6!

I plan to further investigate when time permits, but I'm fine if you want to merge this.

Issue #811, opened 2016-03-02
Geom.bar - Colorkey does not adhere to the order specified in Scale.x_discrete


dstoeckel:
To group related classes together in a Geom.bar plot I used Scale.x_discrete.
The (simplified) code reads as follows:

```julia
df = DataFrame(value=[5,1,6,3], class=["a", "b", "c", "d"]);
plot(
    df, x=:class, y=:value, color=:class,
    Geom.bar, Scale.x_discrete(levels=["b","a","c","d"])
);
```
![gsea1](https://cloud.githubusercontent.com/assets/912333/13457252/6b3ad000-e067-11e5-871c-1fb1bbbcb366.png)

The problem here is, that the colors and the position of the levels in the colorkey are not chosen consistently with the order specified in `levels`. This becomes a problem when multiple plots are generated where new classes are added. In this case I would like to keep a consistent ordering and coloring scheme between the plots by appending the new class to the `levels`. Currently, however, the classnames seem to be sorted lexicographically which will result in shift of all colors.

Mattriks:
In Gadfly (and ggplot2), each aesthetic (e.g. `x`,`y`,`color`) has its own scale, and the scales operate independently.
In Gadfly, the `levels` argument controls the order for each scale:
```Julia
using Gadfly, RDatasets
mtcars = dataset("datasets","mtcars")
p = plot(mtcars, x=:Cyl, color=:Cyl, Geom.histogram,
  Scale.x_discrete(levels=[8,6,4]), Scale.color_discrete(levels=[8,6,4]) )    
```
![scale_discrete](https://user-images.githubusercontent.com/18226881/51800720-762b7680-2287-11e9-8b06-51696292d60b.png)

Mattriks:
In ggplot2, the `limits` argument controls the order:
```R
require("ggplot2")
ggplot(mtcars, aes(x=factor(cyl), fill=factor(cyl))) + geom_bar() + 
 scale_x_discrete(limits=c("8","6","4")) + scale_fill_hue(limits=c("8","6","4"))
```

PR #1245, opened 2019-01-28
Expand Scale sections in the Tutorial

### LINKED ISSUE #573: Trouble using Scale.continuous_scale_partial with the color aesthetic
When plotting bivariate histograms it is often desirable to use a log color scale so most of the dynamic range of the color scale is not used by a few of the bins. As an example, lets look at the "diamond" data-set and look at the carats vs price distribution:

``` julia
using Lazy, Gadfly, RDatasets

d = dataset("ggplot2", "diamonds")

plot(d,
x=:Price,
y=:Carat,
Geom.hexbin(xbincount=70, ybincount=70)) |>
x->draw(SVG(15cm, 15cm), x)
```

produces:
screen shot 2015-03-29 at 8 10 09 pm

Much of the structure of the long tail is hidden as the dynamic range is dominated by a few of the bins in the lower left of the plot. There is no Scale.color_log10, so using the definition as Scale.y_log10 as a rubric, I attempted to roll one myself:

``` julia
color_log10 = Scale.continuous_scale_partial([:color], Scale.log10_transform) # define the scale

plot(d,
x=:Price,
y=:Carat,
color_log10, # use the new scale
Geom.hexbin(xbincount=60, ybincount=60)) |>
x->draw(PNG(15cm, 15cm), x)
```

This gives a somewhat cryptic error:

```
HexBinGeometry requires a continuous color scale.
while loading In[3], in expression starting on line 3
```

Especially given that color_log10 does seem to be a ContinuousScale constructor. Is this not the correct approach to modify color scales? I know that there are methods to produce entirely new scales (like Scale.ContinuousColorScale) but it is hard to make a sensible color scale when them (let alone one that is not linearly scaled). If there is a correct approach to this sort of problem it would be great to make it more accessible / documented.

Mattriks:
- [x] I've updated the documentation

### This PR:
- Expands the sections on Continuous and Discrete Scales in the [Tutorial](http://gadflyjl.org/stable/tutorial/#Scale-transforms-1)
- Inspired by #573 and #811
- closes #573 

Note I leave #811 open, in case anyone wants to implement auto-updating `Scale.color_discrete()` if discrete aesthetic `x`==`color`.

### DIFF
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -1,5 +1,5 @@
 ```@meta
-Author = "Tamas Nagy, Daniel C. Jones, Simon Leblanc"
+Author = "Tamas Nagy, Daniel C. Jones, Simon Leblanc, Mattriks"
 ```
 
 # Tutorial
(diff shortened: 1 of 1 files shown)


tlnagy:
> Note I leave #811 open, in case anyone wants to implement auto-updating Scale.color_discrete() if discrete aesthetic x==color.

I think it's reasonable to assume that if a discrete scale is chosen for `x` and `color` happens to have a 1 to 1 mapping to `x` then to assume the user also wants a `color_discrete` with the same parameters as `x_discrete`

Mattriks:
I agree. Note that functionality is not implemented in ggplot2, as in my [example](https://github.com/GiovineItalia/Gadfly.jl/issues/811#issuecomment-457912004), so it would be nice to have in Gadfly!

This present PR is just about Scale documentation - its ready to merge.

tlnagy:
> Note that functionality is not implemented in ggplot2

It's okay for us to pass `ggplot2` thanks to all your hard work :)

PR #1252, opened 2019-02-16
Scale.alpha_continuous and Scale.alpha_discrete

Mattriks:
- [x] I've updated the documentation to reflect these changes
- [x] I've added an entry to `NEWS.md`
- [x] I've added and/or updated the unit tests
- [x] I've run the regression tests
- [x] I've built the docs and confirmed these changes don't cause new errors

### This PR:
- Adds `Scale.alpha_continuous()` and `Scale.alpha_discrete()`
- `Theme(alphas=[])`
- Basic alpha integration with Guides
- fixes #1226, fixes #1242
- Updated docs

So it is no longer necessary to use `Theme(lowlight_color=)` for opacity. But for example `Theme(lowlight_color=c->"gray")` is still useful for changing the line color of `Geom.ribbon` and `Geom.polygon` independently of the `color` aesthetic.

### Examples:
```julia
palettef = Scale.lab_gradient("darkgreen","orange", "blue")
p1 = plot(x=1:10, y=rand(10), color=[1:10;], Geom.point,  
    Scale.color_continuous(colormap=palettef, minvalue=0, maxvalue=10),
    Guide.title("Scale.color_continuous, Theme(alphas=[0.5])"),
    Theme(alphas=[0.5], continuous_highlight_color=identity, point_size=2mm)
)
p2 = plot(x=1:10, y=rand(10), alpha=[1:10;], Geom.point,  
    Scale.alpha_continuous(minvalue=0, maxvalue=10),
    Guide.title("Scale.alpha_continuous, Theme(default_color=\"blue\")"),
    Theme(default_color="blue", discrete_highlight_color=c->"gray", point_size=2mm)
)
hstack(p1, p2)
```
alpha_continuous

```julia
D = DataFrame(x=1:6, y=[0.39, 0.26, 0.31, 0.43, 0.1, 0.78], Shape=repeat(["a","b","c"], outer=2))
coord = Coord.cartesian(xmin=0, xmax=7, ymin=0, ymax=1.0)
p1 = plot(D, x=:x, y=:y, color=:x,  coord,
    Scale.color_discrete, Geom.point, Geom.hair,
    Guide.title("Scale.color_discrete, Theme(alphas=[0.5])"),
    Theme(alphas=[0.5], discrete_highlight_color=identity, point_size=2mm)
)
p2 = plot(D, x=:x, y=:y, alpha=:x, shape=:Shape, coord,
    Scale.alpha_discrete, Geom.point, Geom.hair,
[...]

PR #1257, opened 2019-03-03
Scale.alpha_continuous & Scale.alpha_discrete

### LINKED ISSUE #1226: opacity in the color scale
this fails:

```
colorx = RGBA(0.0,0.5,0,0.1)
colory = RGBA(0,0.5,0,1.0)
plot(x=[1,2,3], y=[2,6,4], Geom.point(),  Scale.color_continuous(colormap=Scale.lab_gradient(colorx, colory)))

ERROR: StackOverflowError:
Stacktrace:
 [1] lab_gradient(::ColorTypes.RGBA{Float64}, ::Vararg{ColorTypes.RGBA{Float64},N} where N) at /Users/monty/.julia/v0.6/Gadfly/src/color_misc.jl:71 (repeats 20947 times)
```

How i can make it work?

### LINKED ISSUE #1242: svgjs, hstack and opacity
In the v1.0.1 docs, the left plot for [`Geom.ribbon`](http://gadflyjl.org/stable/gallery/geometries/#[Geom.ribbon](@ref)-1) currently has a non-transparent ribbon (mentioned in #1241). I can reproduce this by doing:
```julia
# p1 and p2 from the docs Geom.ribbon example
draw(SVGJS("Docs_ribbon.svg"), hstack(p1,p2))
```
SVGJS seems to be the only backend with this issue; the issue didn't occur for me using other backends. SVG is fine.

Mattriks:

Same as #1252, but now with `alpha` info in the docstrings for `Geom.point`, `Geom.polygon` and `Geom.ribbon` 

- fixes #1226, fixes #1242

### DIFF
--- a/src/geom/polygon.jl
+++ b/src/geom/polygon.jl
@@ -13,7 +13,7 @@ PolygonGeometry(default_statistic=Gadfly.Stat.identity(); order=0, fill=false, p
     Geom.polygon[(; order=0, fill=false, preserve_order=false)]
 
 Draw polygons with vertices specified by the `x` and `y` aesthetics.
-Optionally plot multiple polygons according to the `group`, `color` and/or `linestyle`
+Optionally plot multiple polygons according to the `group`, `color`, `linestyle`, and/or `alpha`
 aesthetics.  `order` controls whether the polygon(s) are underneath or on top
 of other forms.  If `fill=true`, fill the polygons using `Theme.lowlight_color` and stroke the polygons using
 `Theme.discrete_highlight_color`. If `fill=false` stroke the polygons using `Theme.lowlight_color` and `Theme.line_style`.
@@ -22,7 +22,7 @@ around their centroid.
 """
 const polygon = PolygonGeometry
 
-element_aesthetics(::PolygonGeometry) = [:x, :y, :color, :group, :linestyle]
+element_aesthetics(::PolygonGeometry) = [:x, :y, :color, :group, :linestyle, :alpha]
 
 """
     Geom.ellipse[(; distribution=MvNormal, levels=[0.95], nsegments=51, fill=false)]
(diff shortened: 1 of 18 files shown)


src/scale.jl
```diff
@@ -313,7 +325,7 @@ anything in the data that's not respresented in `levels` will be set to
 default order.
 
 See also [`group_discrete`](@ref), [`shape_discrete`](@ref), 
-[`size_discrete`](@ref), and [`linestyle_discrete`](@ref).
+[`size_discrete`](@ref), and [`linestyle_discrete`](@ref), and [`alpha_discrete`](@ref).
```
bjarthur: the "and" before linestyle_discrete could be removed as it was in #1252

src/geom/polygon.jl
```diff
@@ -69,20 +70,25 @@ function render(geom::PolygonGeometry, theme::Gadfly.Theme, aes::Gadfly.Aestheti
...
+    alphas = Vector{Float64}(undef, n)
+    alpha_discrete  = AT <: Int
+
     for (k,g) in enumerate(ug)
         i = groups.==[g]
         polys[k] = polygon_points(aes_x[i], aes_y[i], geom.preserve_order)
         colors[k] = first(aes_color[i])
-        line_styles[k] = mod1(first(aes_linestyle[i]), linestyle_palette_length) 
+        line_styles[k] = mod1(first(aes_linestyle[i]), linestyle_palette_length)
+        alphas[k] = first(alpha_discrete ? theme.alphas[aes_alpha[i]] : aes_alpha[i])
```
bjarthur: what git weirdness is causing the `line_style[k]` line to be highlight as a change?

src/geom/ribbon.jl
```diff
@@ -10,41 +10,42 @@ RibbonGeometry(default_statistic=Gadfly.Stat.identity(); fill=true, tag=empty_ta
     Geom.ribbon
 
 Draw a ribbon at the positions in `x` bounded above and below by `ymax` and
-`ymin`, respectively.  Optionally draw multiple ribbons by grouping with `color`.
+`ymin`, respectively.  Optionally draw multiple ribbons by grouping with `color` and `alpha` or `linestyle`.
```
bjarthur: what does "color and alpha or linestyle" mean?  the boolean logic is unclear

test/testscripts/scale_alpha_discrete.jl
```diff
@@ -0,0 +1,19 @@
...
+    Theme(alphas=[0.5], discrete_highlight_color=identity,
+        point_size=2mm)
+)
+p2 = plot(D, x=:x, y=:y, alpha=:x, shape=:Shape, coord,
+    Scale.alpha_discrete, Geom.point, Geom.hair,
+    Guide.title("Scale.alpha_discrete, Theme(default_color=\"green\")"),
+    Theme(default_color="green",   discrete_highlight_color=c->"gray",
+        point_size=2mm, alphas=[0.0,0.2,0.4,0.6,0.8,1.0])
+)
+hstack(p1,p2)
```
bjarthur: both of these tests are near identical to the docs, the only difference being the hard-coded random numbers.  seems redundant.  ideally i like to see real data, say from RDatasets, used in the docs that is particularly pretty, and very simple made up cases for the tests with focus on execution speed (we time out sometimes) and code coverage.  this is a bit of a nit-pick, but can you at least make the docs and the tests less redundant?

src/scale.jl
Mattriks: Done

src/geom/ribbon.jl
Mattriks: I've clarified this. It differs whether the ribbon is filled or not filled:
https://github.com/GiovineItalia/Gadfly.jl/blob/99f54e467e92c6478dea5ef3ec68ae4bfe30377e/src/geom/ribbon.jl#L75-L76

test/testscripts/scale_alpha_discrete.jl
Mattriks: I like this example because it illustrates e.g. "color aesthetic, fixed alpha" and "alpha aesthetic, fixed color".  If I used an example from RDatasets to illustrate the `alpha` aesthetic, I would use something like the `Diamonds` dataset (53940 points).  In my next PR I'm going to halve the time it takes to plot that: `Geom.point` currently suffers from using a new context for every point.

src/geom/polygon.jl
Mattriks: Weird - no idea.

PR #1263, opened 2019-03-24
feature/rect

CiaranOMara:
Develops `RectangularGeometry` separately from `RectangularBinGeometry`. It was suggested in #1176 that `RectangularGeometry` could be a base geometry.

# Contributor checklist:

# Proposed changes
- Develops `RectangularGeometry` separately from `RectangularBinGeometry`.

# Breaking changes
- Geometry `Geom.rect` no longer reduces height or width by `theme.bar_spacing`. This change leads to a regression in the `Geom.rectbin` test where `Geom.rect` is used.


bjarthur:
great to have this revived!

would you care to take the extra step of deleting `function render(geom::RectangularBinGeometry...` and refactoring `Stat.histogram2d` and `Stat.rectbin` to use `Geom.RectangularGeometry`?

src/geom/rectbin.jl
```diff
@@ -1,13 +1,7 @@
...
-    default_statistic::Gadfly.StatisticElement
-    tag::Symbol
-end
-
 function RectangularBinGeometry(
         default_statistic::Gadfly.StatisticElement=Gadfly.Stat.rectbin();
         tag=empty_tag)
-    RectangularBinGeometry(default_statistic, tag)
+    RectangularGeometry(default_statistic, tag)
 end
```
bjarthur: is there a reason to not delete `RectangularBinGeometry` too?  `rectbin` could then simply be defined as `RectagularyGeometry(Stat.rectbin())`.

bjarthur:
re. the breaking change with `Theme.bar_spacing`...

what do you think of adding two fields to `RectangularGeometry`:  `vpadding::Union{Symbol,Measure}` and  a similarly typed`hpadding`, both of which default to the symbol `:theme_bar_spacing`.  then in `render`, if they are equal to `:theme_bar_spacing` you subtract `theme.bar_spacing`, otherwise you subtract the `Measure`d value.

in #1264 you could them define `band` as a call to `RectangularGeometry` with both paddings set to `0mm`.

would all this work to not break backwards compatibility and to retain the flexibility for bands?

src/geom/rect.jl
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    properties = (
+        stroke(nothing), #TODO: Apply theme attributes.
+        svgclass("geometry"),
+        svgattribute("shape-rendering", "crispEdges"))
+
+    return compose!(
+        context(),
+        Compose.polygon(polys, geom.tag),
```
Mattriks: I ran into issues using polygons where one axis was a date/datetime type in: https://github.com/GiovineItalia/Compose.jl/pull/338#issuecomment-450558055.  You'll need to include tests of that with this PR.

NEWS.md
```diff
@@ -1,6 +1,9 @@
 
 This is a log of major changes in Gadfly between releases. It is not exhaustive.
 Each release typically has a number of minor bug fixes beyond what is listed here.
+# Version 1.x
```
Mattriks: Gadfy hasn't reached 1.1 yet, so this news item can go under the # Version 1.1.0 heading below. And include the (PR number)

Mattriks:
Since this PR updates the `render` function for rectangles, it would be good to add support for the `alpha` aesthetic, so rectangles can work with different `color` and `alpha` aesthetic groupings. `PointGeometry`,  `PolygonGeometry`, and `RibbonGeometry` all currently support `alpha`, so you can look at their `render` functions for examples of how to implement.

src/geom/rectbin.jl
CiaranOMara: For the moment, just the `theme_bar_spacing`. I recognised the same possibility and tried to handle the `theme_bar_spacing` in `Stat.rectbin`, but got knocked back with an incompatible type issue.

CiaranOMara:
@bjarthur, a counter to your reply regarding `theme_bar_spacing`.

What about adding a hook that gets called between the application of statistics and rendering geometries?

In this case the hook signature could be something like `apply_theme_transformations!(geom::RectangularGeometry, stat::RectbinStatistic, aes::Gadfly.Aesthetics, theme::Gadfly.Theme)`.

src/geom/rectbin.jl
CiaranOMara: `RectangularBinGeometry` deleted in https://github.com/GiovineItalia/Gadfly.jl/pull/1263/commits/2c5823054be5581fa74a349828b2973ce1d06a88. This will likely get squashed into another commit.

src/Gadfly.jl
```diff
@@ -1102,6 +1114,17 @@ include("geometry.jl")
...
+#     @info "apply_theme_transformations!" geom stat aes theme
+end
+
+function apply_theme_transformations!(geom::Geom.RectangularGeometry, stat::Stat.RectbinStatistic, aes::Aesthetics, theme::Theme)
+
+    @info "apply_theme_transformations! specific to geom::Geom.RectangularGeometry, stat::Stat.RectbinStatistic combination."
+
+	 aes.xmax = aes.xmax .* cx .- theme.bar_spacing
+	 aes.ymax = aes.ymax .* cy .- theme.bar_spacing
+end
```
bjarthur: @Mattriks @tlnagy thoughts on this proposed new `apply_theme_transformations!` addition to the existing framework?

seems okay to me, but it's not clear what advantages it confers to just handling this case in the corresponding `render` function.

are there other cases which could be refactored like this?

certainly the `@info` and `@warn` need to go.

CiaranOMara: I naively wondered if it were possible to handle default colour and alpha with this sort of pattern.

Mattriks: I'm not liking `apply_theme_transformations!`, starting with the strange looping in `render_prepared()`.    Although `render_prepared` has an argument `layer_stats`, that argument is not used in `render_prepared`, and there is no reason for it to be: the applying of `Stats` is done by `render_prepare()`.

This sort of "code to handle common cases" should be done once there is a good idea of what those common cases are.  For example, over time the `default_aes` list (which appears at the start of render functions for polygon, ribbon, point etc) has become more similar, and this is where "unifying code" might be useful.

CiaranOMara: Regarding the strange looping in `render_prepared()`, I picked a spot between application of statistics and rendering where all of the variables required to demonstrate the concept were readily available; it was by no means a suggestion of final implementation. Commit https://github.com/GiovineItalia/Gadfly.jl/commit/bac3a29bd13ca6cacc45e6fcd22a690a3026fd31 places the `apply_theme_transformations!` methods where I'd expect to find them. However, the commit does not develop the hook call.

Mattriks: In developing a new `apply_` function, I would like to see a design outline like this:

```julia
# Minimal Gadfly

# function render_prepare
datas = Gadfly.Data(x=1:10, y=rand(10), xend=1:10, yend=zeros(10))
scales=Dict{Symbol,Gadfly.ScaleElement}(:x=>Scale.x_continuous(), :y=>Scale.y_continuous())
aes = Scale.apply_scales(collect(values(scales)), datas)
coord = Coord.cartesian()
stats = Gadfly.StatisticElement[Stat.xticks(), Stat.yticks()]
Stat.apply_statistics(stats, scales, coord, aes[1])
# stats have been implemented, and the stats vector is no longer needed

# function render_prepared
plot_context = Coord.apply_coordinate(coord, aes, scales)
layer_themes = [Theme(default_color="green"), Theme(default_color="orange")]
geom_ctxs = render.([Geom.point(), Geom.hair()], layer_themes, aes)
guide_ctxs = render.([Guide.xticks(), Guide.yticks()], layer_themes, aes)
compose!(plot_context, (context(), geom_ctxs...))
tbl =  Guide.layout_guides(plot_context, coord, Theme(), [x[1] for x in  guide_ctxs]...)

p = compose!(context(), tbl)
```
with an explanation about why I want to add a new `apply_` function to the above code. Then I would ask about timings. Also, I would prefer to see such a new function in a separate PR.

CiaranOMara: I will withdraw the commits for the `apply_theme_transformations!` concept and place them in [feature/rect-apply_theme_transformations](https://github.com/CiaranOMara/Gadfly.jl/tree/feature/rect-apply_theme_transformations). When I have the opportunity, I'll open another PR with a satisfactory explanation and design.

src/geom/rect.jl
bjarthur: @Mattriks could you please elaborate on this request for a test?

src/Gadfly.jl
bjarthur: agreed that a major change like that should go in a separate PR.  ashame though to have so much code duplication now between render(RectGeo and render(RectBinGeo.

do the regression tests still pass with things as they stand now?

i'd like to see all four commits squashed into a single one.  no need to clutter the git log like this.

Mattriks: I think it's better for those two geometries (RectGeo and RectBinGeo) to evolve separately (let's give them the chance to do that).

CiaranOMara: @bjarthur, as of the rollback to https://github.com/GiovineItalia/Gadfly.jl/commit/83399184eeb53924d86546472be67f4f8a790fee there is a regression in `testscripts/rectbin.jl` where `Geom.rect` and `Geom.rectbin` intertwine. I think `Geom.rect` should be removed from the `rectbin` test. I'll have a look at the documentation and make sure the geometries are separated there too. However, this regression alerts me to an important question: what is an appropriate overlap between a `max` and `next min` that share the same value? In the current implementation, is it the stroke that overlaps and not the fill (which I think is appropriate), or does the fill overlap? I'm not sure of the specifics, but suppose it depends on `Compose.jl` and the various backends. I can foresee us encountering undesired artifacts when applying alpha to `Geom.rect` if the nuance of overlap is not correct.

Happy to squash commits once everything looks good.

@Mattriks, the rollback to undoes the changes to `Geom.rectbin` to allow it to evolve separately.

src/geom/rectbin.jl
CiaranOMara: Refactoring was reverted to allow `RectangularBinGeometry` and `RectangularGeometry` to evolve separately.

src/Gadfly.jl
bjarthur: is the regression whether the fill and stroke overlap or not?  is there no way to fix it?

CiaranOMara: @bjarthur, I'm confident the regression is due to the exclusion of `theme.bar_spacing` from the new `Geom.rect` implementation.

Below is the `rect.png` test output from https://github.com/GiovineItalia/Gadfly.jl/pull/1263/commits/beee87012e73244ee7be34d6218ad51c508c466a, which is a little fuzzy about the edges.
rect

Below is the `rect.pdf` test output zoomed in on (0.5, 0.5) of the plot on the left side.
image

Given these new images, I do not think there is any overlap of fill in the current implementation. However, stroke is currently `stroke(nothing)`.

In terms of regressions, they're isolated to `Geom.rect`. No other geometries derive from `Geom.rect`, other geometries are deriving from `Geom.rectbin`.

Do you think the exclusion of `theme.bar_spacing` from the new `Geom.rect` makes sense?

bjarthur: i don't think we can tolerate any regressions until we decide it's time to release 2.0.  if you want to keep this as is, we'll have to put a milestone on it and save it for later.

having a way to slightly adjust the dimensions of the rectangles seems generally useful to me. Theme.bar_spacing doesn't seem the right way to go about it though since that sounds specific to Geom.bar.  but i'd suggest sticking with it for now, and perhaps filing an issue with a milestone 2.0 tag to remind us to change it later.

PR #1276, opened 2019-04-17
Remove usage of fieldnames function at runtime where possible

non-Jedi:
Introspection using fieldnames at runtime can have a large negative
effect on code speed.

For time-to-first-plot, which is all I'm interested in at the moment,
these changes give a 5-10% improvement on simple cases I tested.

Before:

```julia
const x = collect(1:2000)
const y = rand(2000)

@time p = plot(x=x, y=y, Geom.line);
 0.637450 seconds (947.25 k allocations: 48.127 MiB, 5.71% gc time)

@time draw(SVG("test.svg", 4cm, 4cm), p)
 60.589051 seconds (82.63 M allocations: 4.113 GiB, 5.98% gc time)
```

After:

```julia
const x = collect(1:2000)
const y = rand(2000)

@time p = plot(x=x, y=y, Geom.line);
 0.594075 seconds (941.17 k allocations: 47.927 MiB, 4.30% gc time)

@time draw(SVG("test.svg", 4cm, 4cm), p)
 54.662001 seconds (81.76 M allocations: 4.070 GiB, 5.74% gc time)
```

I didn't always get the full 10% increase shown here, but it was
consistently better across a few runs.

### DIFF
--- a/src/aesthetics.jl
+++ b/src/aesthetics.jl
@@ -93,11 +93,14 @@ const NumericalAesthetic =
     pad_categorical_y, Union{Missing,Bool}, missing
 end
 
+# Calculating fieldnames at runtime is expensive
+const valid_aesthetics = fieldnames(Aesthetics)
+
 
 function show(io::IO, data::Aesthetics)
     maxlen = 0
     print(io, "Aesthetics(")
-    for name in fieldnames(Aesthetics)
+    for name in valid_aesthetics
         val = getfield(data, name)
         if !ismissing(val) && val != nothing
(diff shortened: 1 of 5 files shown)


bjarthur:
lgtm.  passes regression tests i presume?

non-Jedi:
I knew I was forgetting something. Hold on.

non-Jedi:
Regression tests are okay. I actually got a positive, and then I
realized that the output of `date_bar.jl` depends on the day you run
it, and I was comparing against running the regression suite on master
yesterday. Will open a PR to make that one deterministic.

PR #1277, opened 2019-04-17
Add Project.toml for new Julia Registrator package flow

non-Jedi:
Since packages new register directly to the General repo, a
`Project.toml` will be needed before Gadfly's next release.


bjarthur:
are we supposed to [delete REQUIRE](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/17?u=bjarthur) too?

isn't there supposed to be a [version number](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/32?u=bjarthur) in Project.toml?  cf [1](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/34?u=bjarthur), [2](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/35?u=bjarthur), [3](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/36?u=bjarthur), [4](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/55?u=bjarthur), [5](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/125?u=bjarthur)

did you use [this tool](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/28?u=bjarthur) to generate it? or [this one](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/106?u=bjarthur)?

non-Jedi:
- REQUIRE can certainly be deleted; you're right.
- I did not add a version number because I'm not sure if next release will be 1.1.0 or 1.0.1. One will be needed before next release.
- I don't think I used a tool. This is the Project.toml that's been sitting around on my computer for the past few months for practical reasons in the Gadfly repository; pretty sure I just generated it by hand so that I could have a clean environment for Gadfly development.

non-Jedi:
The other thing we should talk about if whether we want to start
following the practice of doing some rudimentary dependency
pinning. Since Gadfly is post-1.0 now it seems like it shouldn't just
be assumed that either we're compatible with all versions of a package
or that we're compatible with all versions greater than some
number. Should we cap version numbers to current minor version for
pre-1.0 packages and current major version for post? e.g.

```toml
[compat]
Colors = "0.3.4, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9"
```

bjarthur:
there are new features, not just bug fixes, so let's got with 1.1.

should add test/Project.toml too and delete test/REQUIRE

version pinning seem reasonable.  [this script](https://discourse.julialang.org/t/switching-package-registration-systems-soon/22677/106?u=bjarthur) might do that.  would be good to compare it's output with yours anyway.

non-Jedi:
That script generated a file identical to mine with the exception that
it has a different uuid for the `Statistics` package; which makes me
wonder if mine is correct. Could you run the tests using the
Project.toml from this PR and confirm that it works for you?

```julia
cd("Gadfly.jl")
import Pkg
Pkg.activate(".")
Pkg.instantiate()
Pkg.test()
```

non-Jedi:
I've moved test/REQUIRE into the Project.toml file under `[targets]` and `[extras]` as described in the Pkg.jl manual. I also added Julia 0.7 and Julia 1.1 to the CI tests since we're claiming to support them. Versions have been pinned (manually; the script did *not* pin versions). I think this is ready to go.

bjarthur:
looks great.  two things:

could you please squash your first two commits so as not to clutter the git log.

your uuid for Statistics matches mine and the one in the [julia stdlib](https://github.com/JuliaLang/julia/blob/master/stdlib/Statistics/Project.toml).  would be good though to get a firm answer on why @StefanKarpinski 's script is returning a different one.  perhaps submit an issue to Pkg.jl?

non-Jedi:
Done. `Statistics` uuid issue was corrected since I downloaded the script: https://github.com/JuliaLang/Pkg.jl/pull/1138.

Issue #1275, opened 2019-04-17
Gadfly first-plot profiling results


non-Jedi:
I got curious about "time to first plot" and did some profiling to see
where most of the problem was coming from. I thought it might be
useful to share the results here for other's reference so that you
don't have to figure out the proper incantations of `Profile`
yourself. The approximate recipe for reproducing this is:

```julia
using Gadfly, Profile

x = collect(1:2000)
y = rand(2000)
p = plot(x=x, y=y, Geom.line);
Profile.init(10000000, 0.005)
@profile draw(SVG("test.svg", 4cm, 4cm), p)
```

Here's the results of the profiling in tree-view:

```
8277 /home/adam/repos/Gadfly.jl/src/Gadfly.jl:851; draw(::SVG, ::Plot)
 1982 /home/adam/.julia/packages/Compose/aVZwV/src/container.jl:334; draw
  1965 /home/adam/.julia/packages/Compose/aVZwV/src/container.jl:487; drawpart(::SVG, ::Compose.Context, ::Compose.IdentityTransform, ::Compose.UnitBox{Float64,Float64,Float64,Float64}, ::Measures.BoundingBox{Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}},Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}}})
   1791 /home/adam/.julia/packages/Compose/aVZwV/src/container.jl:487; drawpart(::SVG, ::Compose.Context, ::Compose.IdentityTransform, ::Compose.UnitBox{Float64,Float64,Float64,Float64}, ::Measures.BoundingBox{Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}},Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}}})
    1463 /home/adam/.julia/packages/Compose/aVZwV/src/container.jl:368; drawpart(::SVG, ::Compose.Table, ::Compose.IdentityTransform, ::Compose.UnitBox{Float64,Float64,Float64,Float64}, ::Measures.BoundingBox{Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}},Tuple{Measures.Length{:mm,Float64},Measures.Length{:mm,Float64}}})
[...]

PR #1278, opened 2019-04-19
WIP: Time-to-first-plot improvements

non-Jedi:
Fixes #XXX

These two changes give close to a 20% speedup in time-to-first-plot in my informal bench-marking. Unfortunately, this produces different output on some of the regression suite (thus the WIP); things aren't getting drawn in the same order as before (see below). A total of 17 images are different. Don't worry; I'll figure it out.

density2d

density2d

The changes here are just my fumbling attempts to address the profiling in #1275. I honestly expected the first commit in this pack to have a bigger effect than it did, and I expected the nothing comparisons to be negligible when in reality they were more important than getting rid of the splatting.

# Contributor checklist:

# Proposed changes

-
-
-

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -388,14 +388,14 @@ function render_prepare(plot::Plot)
     # they are missing.
     datas = Array{Data}(undef, length(plot.layers))
     for (i, layer) in enumerate(plot.layers)
-        if layer.data_source === nothing && isempty(layer.mapping)
+        if isnothing(layer.data_source) && isempty(layer.mapping)
             layer.data_source = plot.data_source
             layer.mapping = plot.mapping
             datas[i] = plot.data
         else
             datas[i] = Data()
 
-            if layer.data_source === nothing
+            if isnothing(layer.data_source)
                 layer.data_source = plot.data_source
             end
 
@@ -416,7 +416,7 @@ function render_prepare(plot::Plot)
         if isa(layer.geom, Geom.SubplotGeometry)
             for subplot_layer in layers(layer.geom)
                 subplot_data = Data()
-                if subplot_layer.data_source === nothing
+                if isnothing(subplot_layer.data_source)
                     subplot_layer.data_source = layer.data_source
                 end
 
@@ -434,7 +434,7 @@ function render_prepare(plot::Plot)
     coord = plot.coord
     for layer in plot.layers
         coord_type = element_coordinate_type(layer.geom)
-        if coord === nothing
+        if isnothing(coord)
             coord = coord_type()
         elseif typeof(coord) != coord_type
             error("Plot uses multiple coordinates: $(typeof(coord)) and $(coord_type)")
@@ -503,7 +503,7 @@ function render_prepare(plot::Plot)
 
     unscaled_aesthetics = setdiff(used_aesthetics, scaled_aesthetics)
 
-    _theme(plt, lyr) = lyr.theme == nothing ? plt.theme : lyr.theme
+    _theme(plt, lyr) = isnothing(lyr.theme) ? plt.theme : lyr.theme
 
     # Add default scales for statistics.
     layer_stats_with_theme = map(plot.layers, layer_stats) do l, stats
@@ -532,17 +532,17 @@ function render_prepare(plot::Plot)
         in(var, mapped_aesthetics) || continue
 
         var_data = getfield(plot.data, var)
-        if var_data == nothing
+        if isnothing(var_data)
             for data in datas
                 var_layer_data = getfield(data, var)
-                if var_layer_data != nothing
+                if issomething(var_layer_data)
                     var_data = var_layer_data
                     break
                 end
             end
         end
 
-        var_data == nothing && continue
+        isnothing(var_data) && continue
 
         t = classify_data(var_data)
         if scale_exists(t, var)
@@ -560,7 +560,7 @@ function render_prepare(plot::Plot)
         t = :categorical
         for data in Iterators.flatten((datas, subplot_datas))
             val = getfield(data, var)
-            if val != nothing && val != :categorical
+            if issomething(val) && val != :categorical
                 t = classify_data(val)
             end
         end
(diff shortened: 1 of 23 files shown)


non-Jedi:
Just noticed some other miscellaneous refactoring made it into the first commit other than its stated purpose of eliminating splatting. I think they're all changes for the good, but I can rebase them out if you'd prefer a cleaner review here.

src/geom/point.jl
```diff
@@ -67,7 +67,7 @@ function render(geom::PointGeometry, theme::Gadfly.Theme, aes::Gadfly.Aesthetics
     for (x, y, color, size, shape, alpha) in Compose.cyclezip(aes.x, aes.y, aes.color, aes.size, aes.shape, aes_alpha)
         shapefun = typeof(shape) <: Function ? shape : theme.point_shapes[shape]
         sizeval = typeof(size) <: Int ? interpolate_size(size) : size
-        strokecolor = aes.color_key_continuous != nothing && aes.color_key_continuous ?
+        strokecolor = issomething(aes.color_key_continuous) && aes.color_key_continuous ?
```
Mattriks: Note PR #1258 which rewrites the `render` for PointGeometry.

bjarthur:
thanks for you work on this dire problem!  i'd been assuming it was hopeless from a pkg dev point of view and that only changes to julia base could fix it.

src/geom/point.jl
non-Jedi: How close is that PR to merging? I can either rebase this on top of it, you can rebase on top of this, or we can wait and see whichever gets merged first. I did notice you used `!= nothing` a few times in your rewrite; whatever we do, we should make sure that those are changed to `issomething` before they make it to master.

non-Jedi:
Well, I'm down to 7 that are different now (all but one entirely new from the previous list):

- colorful_hist
- hexbin
- hexbin_dark
- stacked_continuous_histogram
- stacked_discrete_histogram
- stacked_discrete_histogram_horizontal
- subplot_grid_free_y_2

and to be honest for some of them I'm not seeing the difference. Any thoughts? I'll keep working. Examples:

## colorful_hist

No idea what's going on here.

### master

colorful_hist

### devel

colorful_hist

### diff

colorful_hist png

## hexbin_dark

different background color

### master

hexbin_dark

### devel

hexbin_dark

### diff

hexbin_dark png

## stacked_continuous_histogram

no idea

### master

stacked_continuous_histogram

### devel

stacked_continuous_histogram


### diff

stacked_continuous_histogram png


## subplot_grid_free_y_2

Scales are off on this one. Weird that it's the only one that seems to have this problem.

### master

subplot_grid_free_y_2

### devel

subplot_grid_free_y_2

### diff

subplot_grid_free_y_2 png

non-Jedi:
All regression tests now pass. I'm gonna run some stuff to make pretty graphs showing the improvement as is only proper when working on Gadfly.

non-Jedi:
So the results weren't as dramatic for these particular changes running on my desktop versus running on my laptop. Still a nice little ~5% bump though. [I threw the data from running through all the testscripts up on Github](https://github.com/non-Jedi/Gadfly_bench_data) if anyone is curious enough to take a closer look.

drawtime

drawallocs

bjarthur:
lgtm.  4% is better than nothing.

where is `issomething` defined?  i can't find it!

you might consider a PR to the "performance tips" of the julia docs.  there is no mention of `isnothing` being faster and splats being slow.

@Mattriks you ready to merge #1258 ?

non-Jedi:
Both `isnothing` and `issomething` are actually defined in `Gadfly`'s `misc.jl` which is necessary to support 1.0 since `isnothing` wasn't introduced to `Base` until 1.1. Which reminds me: the `Base` implementation uses type dispatch instead of `===`. I should change `misc.jl` to match to see if we can get a tiny bit of additional speedup there.

non-Jedi:
Force pushed with changes mentioned above to `isnothing` and a rebase on master to resolve conflicts with #1276.

src/misc.jl
```diff
@@ -188,7 +188,8 @@ function inherit!(a::T, b::T) where T
 end
 
 
-isnothing(u) = u === nothing
+isnothing(::Nothing) = true
+isnothing(::Any) = false
```
bjarthur: `isnothing` is in Compat now, so we shouldn't need this.  https://github.com/JuliaLang/Compat.jl/pull/654

non-Jedi: To make the code more readable without having to go find where `issomething` is defined, I've replaced all instances of `issomething` with `!isnothing` and removed both from misc.jl. I also updated Project.toml to specify 2.1 for `Compat` so that we know we have `isnothing`.

non-Jedi:
CI failure is unrelated. I'll try to PR a version bump of Documenter in the next few days to fix doc building failure.

bjarthur:
perhaps CI failure for docs is due to forgetting to fold docs/Project.toml into ~/Project.toml?

src/geom/rectbin.jl
```diff
@@ -94,14 +94,14 @@ function render(geom::RectangularBinGeometry, theme::Gadfly.Theme, aes::Gadfly.A
...
-        if c == nothing
+        if isnothing(c)
             allvisible = false
             break
         end
     end
 
     if !allvisible
-        visibility = cs .!= nothing
+        visibility = !isnothing.(cs)
```
bjarthur: shouldn't the `!` need to be broadcast here?  so `visibility  = .!isnothing.(cs)`.

codvec says this line [isn't covered](https://codecov.io/gh/GiovineItalia/Gadfly.jl/pull/1278/src/src/geom/rectbin.jl).  would be great if you could add a test.

non-Jedi: Good catch. Fixed. I'm not grocking that code well enough to understand how to trigger it in a test case unfortunately.

non-Jedi:
docs/Project.toml doesn't have a way to be folded into ./Project.toml unfortunately. The problem is that Documenter 0.20 has a hard version cap on DocumentStringExtensions of 0.2 which conflicts with current requirements in ./Project.toml. Should be fixable by bumping Documenter to 0.22.

src/coord.jl
```diff
@@ -67,7 +67,7 @@ function first_concrete_aesthetic_value(aess::Vector{Gadfly.Aesthetics}, vars::V
     for var in vars
         for aes in aess
             vals = getfield(aes, var)
-            vals === nothing && continue
+            Gadfly.isnothing(vals) && continue
```
bjarthur: qualifying `isnothing` with `Gadfly` is not really needed here, is it? it's done throughout src/coord.jl.

src/Gadfly.jl
```diff
@@ -388,14 +388,14 @@ function render_prepare(plot::Plot)
     # they are missing.
     datas = Array{Data}(undef, length(plot.layers))
     for (i, layer) in enumerate(plot.layers)
-        if layer.data_source === nothing && isempty(layer.mapping)
+        if isnothing(layer.data_source) && isempty(layer.mapping)
```
ararslan: I'd be willing to bet `=== nothing` is more efficient. The compiler loves `===`.

ararslan: Simply changing occurrences of `== nothing` to use `===` (and likewise for the negative) should provide a nice improvement.

src/coord.jl
ararslan: Any binding defined locally already shadows one defined in a loaded module (e.g. Base) unless the loaded binding is resolved first. So unless there is a variable inside of this function called `isnothing`, it does not need qualification.

src/Gadfly.jl
ararslan: For reference, this issue applies: https://github.com/JuliaLang/julia/issues/27681

non-Jedi: Good to know. I'll update with straight `===` and see if it makes a difference.

src/coord.jl
non-Jedi: This is all within the module `Gadfly.Coord`. `using Compat` is only within the parent `Gadfly` module, so to use `isnothing` here, we either have to add `using Compat` to this module or qualify usage of `isnothing` with `Gadfly.isnothing`. Stylistically, probably better to just add `using Compat` I suppose.

src/Gadfly.jl
ararslan: That would also resolve the `isnothing` name qualification discussion.

non-Jedi: I've pushed a commit changing from `isnothing` to `=== nothing`. Still need to get back to the computer I've benchmarked on before to see if it makes a difference.

src/coord.jl
non-Jedi: Resolved by switching from `isnothing` to `=== nothing`

PR #1279, opened 2019-04-19
Make testscript date_bar.jl deterministic wrt date run

non-Jedi:
Non-deterministic testscript discovered while regression testing #1276.

# Contributor checklist:


bjarthur:
thanks!

PR #1295, opened 2019-07-02
Accept ranges in xticks, yticks

### LINKED ISSUE #1293: Guide.ytick should accept a range
```julia
using Statistics
using Gadfly

plot(
    y=1:15|> cumsum,
    Geom.line, 
    Guide.yticks(ticks = 0:25:100 |> collect))
```

It would be nice not to have to collect the ticks myself.

In my testing adding this constructor is sufficient:

```julia
Yticks(label, ticks::AbstractRange, orientation) =
    Yticks(label, collect(ticks), orientation)
```

cmcaine:
Fixes #1293

# Contributor checklist:


bjarthur:
sorry i didn't chime in early here.  i think a better approach would be to modify `text()` in Compose.jl to accept a `Range`.  right here:  https://github.com/GiovineItalia/Compose.jl/blob/master/src/form.jl#L418/.  it is not performant to collect an Array unnecessarily.

cmcaine:
I considered looking deeper for something like that but didn't because I don't know what else would plausibly send a range.

I believe the implementation that I have provided will be performant because:

1. It does not collect arrays, only ranges
2. The ranges will be small (people don't specify enormous ranges for xticks, right?)

If you're sure then I'll have a look at updating the `text` function, though. Looks like @makeform just wants an indexable thing, so ranges could work if I just take some type definitions off. I don't know what that will do to your error readability or performance, though.

cmcaine:
In form.jl, is there a reason that there are two `text()` methods next to each other? The methods differ only in the signature for values, so I would expect the explicit methods to be unnecessary, but I'm still learning Julia.

I also observe that `typeof(1:10) <: AbstractArray`, so maybe `text()` is not the problem?

The exact error I get from providing a range is:

```
› julia -e 'using Gadfly; plot(x=1:10, y=1:10, Guide.xticks(ticks=2:9), Geom.point) |> display'
ERROR: TypeError: in setfield!, expected Union{Nothing, Array{T,2} where T, Array{T,1} where T}, got UnitRange{Int64}
Stacktrace:
 [1] apply_statistic(::Gadfly.Stat.TickStatistic, ::Dict{Symbol,Gadfly.ScaleElement}, ::Gadfly.Coord.Cartesian, ::Gadfly.Aesthetics) at /home/colin/.julia/packages/Gadfly/09PWZ/src/statistics.jl:935
 [2] apply_statistics(::Array{Gadfly.StatisticElement,1}, ::Dict{Symbol,Gadfly.ScaleElement}, ::Gadfly.Coord.Cartesian, ::Gadfly.Aesthetics) at /home/colin/.julia/packages/Gadfly/09PWZ/src/statistics.jl:34
 [3] render_prepare(::Plot) at /home/colin/.julia/packages/Gadfly/09PWZ/src/Gadfly.jl:694
 [4] render(::Plot) at /home/colin/.julia/packages/Gadfly/09PWZ/src/Gadfly.jl:745
 [5] draw at /home/colin/.julia/packages/Gadfly/09PWZ/src/Gadfly.jl:854 [inlined]
 [6] display(::Gadfly.GadflyDisplay, ::MIME{Symbol("text/html")}, ::Plot) at /home/colin/.julia/packages/Gadfly/09PWZ/src/Gadfly.jl:1058
 [7] display(::Gadfly.GadflyDisplay, ::String, ::Any) at ./multimedia.jl:178
 [8] display(::Gadfly.GadflyDisplay, ::Plot) at /home/colin/.julia/packages/Gadfly/09PWZ/src/Gadfly.jl:1015
 [9] display(::Any) at ./multimedia.jl:287
 [10] |>(::Plot, ::typeof(display)) at ./operators.jl:813
 [11] top-level scope at none:0
julia -e   39.05s user 0.53s system 99% cpu 39.862 total
```

bjarthur:
oh, good catch.  i suspect that the `text()` with `AbstractString` is vestigial and could be removed.

based on the line numbers i gather you're not on the master branch of Gadfly.

the `setfield` is on an `Aesthetic`, which is defined [here](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/aesthetics.jl#L56).  and the `NumericalAesthetic` therein is defined [here](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/aesthetics.jl#L7) to be `Vector`.  i suspect that `AbstractVector` instead would work.

thanks for looking into this!

cmcaine:
Updated as you suggested and this seems to work.

I've changed a bunch of examples in the docs that unnecessarily collect ranges as well.

I changed Matrix -> AbstractMatrix as well because I didn't see any reason not to do so.

Regression tests pass for outputs other than PS and PDF and I don't visually see any difference in the PDF output for one randomly selected test.

When developing, it's a bit of a pain that the regression tests take so long to run. It would be nice if that was a lot faster or if the CI bots were to do some of that. Perhaps we could just find a representative subsample of tests to run? And do them in parallel?

NEWS.md
```diff
@@ -2,6 +2,10 @@
 This is a log of major changes in Gadfly between releases. It is not exhaustive.
 Each release typically has a number of minor bug fixes beyond what is listed here.
 
+# Version next
+
+ * Support AbstractVectors everywhere (e.g. `Guide.xticks(ticks=1:10)`) (#1293)
+
```
bjarthur: next version will be 1.1.0, so please move this below.  otherwise, looks great!  thanks again for this.  will merge as soon as this change is made.

cmcaine: Done!

PR #1303, opened 2019-08-04
Fix DataFrame deprecation of df[col]

Moelf:


bjarthur:
to get this PR to pass unit tests Project.toml also needs to be changed to specify `DataFrames = "0.19"`

tlnagy:
Looks like we should also drop 0.7 compatibility (in `.travis.yml` and `Project.toml`) going forward with this PR, given that DataFrames v0.18+ already did that.

bjarthur:
no, because DataFrames is an optional dependency via Requires

tlnagy:
Is it mandatory in the tests then? Because the tests on 0.7 are failing with this change in dependency resolution

bjarthur:
yea, i just noticed that.  it is mandatory in the tests.  so maybe we stop testing on 0.7 (change travis.yml) but leave it as an option in Projects.toml?

or, we ditch DataFrames and re-write to use the generic Tables.jl or TableTraits.jl API?

tlnagy:
Superseded by #1318

PR #1331, opened 2019-10-20
Fix #1330 handle closures with special scaling

### LINKED ISSUE #1330: y Scale options do not work for a list of functions properly
I've got a function that as an input takes a list of values and returns an array of functions that I'd like to plot with Gadfly. Everything works fine until I try to plot it in any of the y-scales, then suddenly no plots are produced.
Here's a MWE:

    using Gadfly
    function test_f(a)
        return [x -> a[i]*x for i in 1:length(a)]
    end
    plot(test_f(1:10),1,2π) |> SVG("test.svg")

produces:

[test.pdf](https://github.com/GiovineItalia/Gadfly.jl/files/3746848/test.pdf)

However,

    plot(test_f(1:10),1,2π,Scale.y_log10()) |> SVG("test.svg")

fails to produce any plots and gives the following error:

> MethodError: no method matching log10(::getfield(Main, Symbol("##10#12")){Int64,UnitRange{Int64}})
Closest candidates are:
  log10(!Matched::BigFloat) at mpfr.jl:671
  log10(!Matched::Missing) at math.jl:1072
  log10(!Matched::Complex{Float16}) at math.jl:1020
  ...

laborg:
Fixes #1330
Relax constraint to subtypes of `Function` to allow closures with scaling.

# Contributor checklist:

- [x] I've run the regression tests (they fail on master...)

### DIFF
--- a/src/scale.jl
+++ b/src/scale.jl
@@ -172,7 +172,7 @@ const slope_continuous = continuous_scale_partial([:slope], identity_transform)
     alpha_continuous[(; minvalue=0.0, maxvalue=1.0, labels=nothing,
                      format=nothing, minticks=2, maxticks=10, scalable=true)]
 
-Rescale the data values between `minvalue` and `maxvalue` to opacity (alpha) values between 0 and 1.  
+Rescale the data values between `minvalue` and `maxvalue` to opacity (alpha) values between 0 and 1.
 """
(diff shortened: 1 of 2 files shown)


bjarthur:
looks great!  thanks for helping out.

Issue #1320, opened 2019-09-18
Cairo and Fontconfig incompatible with Gadfly


wilburtownsend:
I need to use the Cairo and Fontconfig packages to export to png.
But they seem to be incompatible with Gadfly.

Running this code:
```
using Gadfly, Cairo, Fontconfig
plot(sin, 0, 6)
```
causes the error 
```
Error showing value of type Plot:
ERROR: TypeError: in typeassert, expected Compose.PangoLayout, got Nothing
```

I'm using Julia 1.1.1 with:
```
  "Fontconfig"       => v"0.2.0"
  "Gadfly"           => v"1.1.0"
  "Cairo"            => v"0.7.0"
```

Mattriks:
What is your OS?
Here are some things to try:
1) Are you on the master branch of Compose.jl (the graphics package used by Gadfly)?  See previous [issue](https://github.com/GiovineItalia/Gadfly.jl/issues/1291#issuecomment-504768594).  You can get the master branch by doing ```]add Compose#master```. See the [Pkg docs](https://julialang.github.io/Pkg.jl/v1/managing-packages/#Adding-registered-packages-1).
2) Alternatively, you may need to ```]add Cairo@0.6``` because Compose.jl hasn't been updated to [Cairo 0.7](https://github.com/GiovineItalia/Compose.jl/blob/a081070de98b7c5d46bc4a0aaf19426a71489b8d/Project.toml#L20)
3. If 1 and 2 fail, try 
```
import Cairo
using Gadfly
```
without importing or using `Fontconfig`

wilburtownsend:
@Mattriks I'm using  OS X 10.14.6.

Thanks -- the third approach you suggested worked. The first didn't seem to do anything. For the second, after running `]add Cairo@0.6`, I received a `LoadError: Cairo not properly installed. Please run Pkg.build("Cairo") `
After running ` Pkg.build("Cairo")`, I received this error.
```
┌ Error: Error building `Cairo`: 
│ WARNING: redefining constant DEBIAN_VERSION_REGEX
│ WARNING: redefining constant have_sonames
│ WARNING: redefining constant sonames
│ WARNING: redefining constant EXTENSIONS
│ Error: glib: undefined method `uses_from_macos' for Formulary::FormulaNamespacee1bb1d62184cef8bbbb76e27789171ae::Glib:Class
[...]

PR #1339, opened 2019-11-03
Fixes #1338

Mattriks:
- [x] I've added and/or updated the unit tests
- [x] I've run the regression tests
- [x] I've built the docs and confirmed these changes don't cause new errors

### This PR:

- Fixes #1338
- Makes the default behaviour for `Geom.point` consistent with the other discrete aesthetics
- Makes the syntax e.g. `shape=x`, when length(x)==1, work with `Geom.subplot_grid`

### Examples
```julia
# p2 and p3 refer to the 2nd and 3rd plot in #1338
y = rand(10)
df = DataFrame(x=repeat(1:5,2), y=y, z=repeat(["a","b"], inner=5))
stylef(shape::Function) = style(point_shapes=[shape], point_size=2mm)

p2 = plot( layer(x=1:10, y=y, stylef(Shape.cross)),
    layer(x=1:10, y=y.+0.3, stylef(Shape.xcross)) )
p3 = plot(df, xgroup=:z, x=:x, y=:y, shape=[Shape.square], alpha=[0.1],
    Geom.subplot_grid(Geom.point), 
    Theme(point_size=5pt, key_position=:none, discrete_highlight_color=identity)
)
hstack(p2, p3)
```
issue1338

Issue #1324, opened 2019-10-03
ERROR: TypeError: in typeassert, expected Compose.PangoLayout, got Nothing


montyvesselinov:
I have to pin cairo @ 0.6.0 to avoid this error.

Mattriks:
An issue has been opened in [Compose](https://github.com/GiovineItalia/Compose.jl#compose) to update Compose to use Cairo 0.7: https://github.com/GiovineItalia/Compose.jl/issues/357

PR #1340, opened 2019-11-06
CI Update

CiaranOMara:
This pull request was created in response to https://github.com/GiovineItalia/Gadfly.jl/pull/1339#issuecomment-550114393.
# Proposed changes

- ~~Add Julia 1.3 to test matrix.~~
- Use the codecov flag in Travis instead of the manual code version.
- Pkg.instantiate after Pkg.develop.
- Add colour to make doc output.
- Move tests with Compose.jl#master into an Edge stage.

Moving tests using Compose.jl#master into their own stage provides visibility for new development while prioritising Gadfly.jl#master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,4 @@ jobs:
           - julia: 1.2
       before_script:
         - julia --color=yes -e 'using Pkg; Pkg.add(PackageSpec(name="Compose", rev="master"))'
-      script:
-        - julia --color=yes --project --check-bounds=yes -e 'import Pkg; Pkg.instantiate(); Pkg.test("Gadfly")'
       after_success: skip
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,43 @@
 language: julia
+
 os:
   - linux
+
 julia:
   - 1.0
   - 1.2
   - nightly
+
 matrix:
   fast_finish: true
   allow_failures:
     - julia: nightly
+
 notifications:
   email: false
-before_script:
-  - julia -e 'using Pkg; Pkg.add(PackageSpec(name="Compose", rev="master"))'
+
 codecov: true
 
 jobs:
   include:
     - stage: "Documentation"
-      julia: 1.0
+      julia: 1.2
       os: linux
       # disable global before_script in order not to install Compose twice
       before_script:
       script:
-        - |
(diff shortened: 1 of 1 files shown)


CiaranOMara:
I wasn't sure how to best structure this pull request. Anyhow, I've added Cairo.jl compatibility on the back of this PR as it not only shows improvement across operating systems, it also shows how the Edge stage works - you'll notice that many of the deprecation warnings are already addressed by Compose.jl#master.

Mattriks:
I assume this PR will close #1320 and #1324. If so, add that to your first post above (https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords)

bjarthur:
it'd be better to have the Cairo dependency commit in a separate PR.  i'd merge that right away.

regarding 1.3, again, i'd suggest holding off on that until it is actually released, and stop testing on 1.2 at the same time.

i also don't understand the Edge job.  i'm just being dumb here i think.  can you please explain how this is different?

CiaranOMara:
Previously, by way of the `before_script`, all tests were performed using `Compose.jl#master`. With this PR, tests in the **Test job** use a `Compose.jl` release and tests in the **Edge job** use `Compose.jl#master`. With this configuration, the tests in the **Test job** test the packages installed by the end-user, and the tests in the **Edge job** shows developers how `Gadfly.jl` is performing with the latest `Compose.jl` developments.

CiaranOMara:
I have moved the Cairo compatibility change to https://github.com/GiovineItalia/Gadfly.jl/pull/1343.

CiaranOMara:
I have removed Julia v1.3 from the test matrix.

.travis.yml
```diff
@@ -1,34 +1,43 @@
 language: julia
+
 os:
   - linux
+  - osx
+  - windows
```
bjarthur: the thinking in only testing Gadfly on linux is that all the OS-specific code is in Compose.  no need to burden the Travis servers.

CiaranOMara: I did think that to be the case. Though their brief inclusion was useful to see Gadfly working with the new binary provided Cairo. Anyhow, I've removed macOS and Windows from the test matrix.

CiaranOMara:
The Edge stage does not seem quite right as it does not appear to be allowed to fail.

Issue #751, opened 2015-11-19
creating empty plot


joachimdahl:
I am generating sparsity plots as

```
plot(x=[1:10;],y=[1:10;],Guide.xticks(ticks=[1,10]),Guide.yticks(ticks=[1,10]))
```

and sometimes the underlying matrices happen to be zero, and I think this is a corner-case missing from plot(), e.g.,

```
julia> plot(x=[],y=[],Guide.xticks(ticks=[1,10]),Guide.yticks(ticks=[1,10]))
ERROR: BoundsError: attempt to access 0-element Array{Any,1}
  at index [1]
 in first at abstractarray.jl:63
 in apply_statistic at /home/joachim/.julia/v0.4/Gadfly/src/statistics.jl:777
 in apply_statistics at /home/joachim/.julia/v0.4/Gadfly/src/statistics.jl:38
 in render_prepare at /home/joachim/.julia/v0.4/Gadfly/src/Gadfly.jl:676
 in render at /home/joachim/.julia/v0.4/Gadfly/src/Gadfly.jl:717
 in draw at /home/joachim/.julia/v0.4/Gadfly/src/Gadfly.jl:821
```

bjarthur:
a similar corner case which errors out is when `all(isnan,y)==true`

acrosby:
Any progress on this, or a clever work around? Thanks!

kilianbreathnach:
I've managed to use hstack and vstack to make an array of subplots with empty space in place of certain axes with this hack:

```
subplots = Array{Gadfly.Plot, 1}(9)

for i in 1:9

    <...>

    if (<"some condition">)
        subplots[i] = plot(x=[0,1], y=[0,1], Geom.point,
                           Guide.xticks(ticks=nothing), Guide.yticks(ticks=nothing),
                           Guide.xlabel(nothing), Guide.ylabel(nothing),
                           Theme(default_color="white"))
        end

    <...>

end

fig = vstack(hstack(subplots[1:3]), hstack(subplots[4:6]), hstack(subplots[7:9]))

```
So basically just a plot of white points with no axes or labels.

bjarthur:
i think the original poster was hoping to have `<some condition>` checked for automatically.  to leave a panel blank, it's easiest to just use `context()`, as described [here](http://gadflyjl.org/latest/man/layers.html#Stacks-1).

PR #1343, opened 2019-11-09
Set Cairo compatibility

CiaranOMara:
This PR resolves dependencies for #1339, closes #1320 and also closes #1324.

PR #1345, opened 2019-11-12
Geom.blank

### LINKED ISSUE #947: Zero-argument function definition missing for plot()
`plot()`
returns: 
`ERROR: The following aesthetics are required by Geom.point but are not defined: y, x` 

This is an old issue, mentioned [here](https://github.com/tbreloff/Plots.jl/issues/73), when Gadfly was used as a backed to Plots. What is the current work around?

What I really want to do is plot circles where the radius is in the same coordinates as the position of the circles. Something that can not easily be done in the package Plots. That is,

`using Gadfly, Compose`
`pc = Guide.annotation(compose(context(), circle([2,4], [2,4], [sqrt(8)/2])))`
`plot(pc)`

One solution is to specify `y,x`:

`p1=layer(x=1:10,y=1:10, Geom.point())`
`plot(p1,ps, Coord.cartesian(fixed=true))`

where `Coord.cartesian(fixed=true)` is passed so that the circles have the correct dimension.

Gadfly is a great package, so thanks to all who help develop it! So I'm wondering, several of these issues are from around 1 year ago, has development slowed down?

Mattriks:

### This PR
- Adds `Geom.blank` 
- Adds `plot()` (closes #947) which is user-friendly, and allows `push!`
- Adds docstring for `push!`, plus doc examples
- Is not aimed at the case with explicit empty arrays: `x=[], y=[]` (#751), which should be handled separately

### Example
```julia
p1 = plot(x=[0,10], y=[0,1], Geom.blank)
p2 = plot()
p3 = push!(plot(), layer(x=1:5, y=rand(5), Geom.point))
hstack(p1, p2, p3)
```
geom_blank

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,8 @@ This is a log of major changes in Gadfly between releases. It is not exhaustive.
 Each release typically has a number of minor bug fixes beyond what is listed here.
 
 # Version 1.x
-* Support DataFrames.jl 0.19 changes in indexing (#1318)
+ * Add `Geom.blank` (#1345)
+ * Support DataFrames.jl 0.19 changes in indexing (#1318)
 
 
 
--- a/docs/src/gallery/geometries.md
+++ b/docs/src/gallery/geometries.md
@@ -57,6 +57,15 @@ set_default_plot_size(14cm, 8cm)
 plot(dataset("lattice", "singer"), x="VoicePart", y="Height", Geom.beeswarm)
 ```
 
+## [`Geom.blank`](@ref)
+
+```@example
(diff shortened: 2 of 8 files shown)


bjarthur:
docs are timing out, despite the [recent PR](https://github.com/GiovineItalia/Gadfly.jl/pull/1343/files) adding `travis_wait`.

Mattriks:
This is running travis with the old travis.yml (before #1343). I'll wait til you merge #1347 and #1348, then rebase.

Mattriks:
Done.

PR #1350, opened 2019-11-17
add compat for CategoricalArrays 0.6

bjarthur:
CompatHelper.jl missed this


bjarthur:
tests with CategoricalArrays pinned to 0.6 pass locally, so i'll merge.

Issue #120, opened 2013-11-14
assertion failed: :((len>0))


Keno:
Hit the following assertion failure:

```
assertion failed: :((len>0))
 in error at error.jl:22
 in format_fixed_scientific at /Users/kfischer/.julia/Gadfly/src/format.jl:74
 in anonymous at /Users/kfischer/.julia/Gadfly/src/format.jl:137
 in identity_formatter at /Users/kfischer/.julia/Gadfly/src/scale.jl:66
 in render at /Users/kfischer/.julia/Gadfly/src/guide.jl:403
 in render_prepared at /Users/kfischer/.julia/Gadfly/src/Gadfly.jl:668
 in render at /Users/kfischer/.julia/Gadfly/src/Gadfly.jl:614
 in writemime at /Users/kfischer/.julia/Gadfly/src/Gadfly.jl:706
 in sprint at io.jl:434
 in display_dict at /Users/kfischer/.julia/IJulia/src/execute_request.jl:36
```

Data is here: https://gist.github.com/loladiro/7463559

```
plot(readtable("test.csv"),x="x1",y="x2")
```

is the failing command.

PR #1351, opened 2019-11-17
remove test dependency on CSV

bjarthur:

only one test used it, so i refactored to use DelimitedFiles in the standard library instead.

### DIFF
--- a/test/testscripts/issue120.jl
+++ b/test/testscripts/issue120.jl
@@ -1,5 +1,8 @@
-using Gadfly, DataFrames, Compat, CSV
+using Gadfly, DataFrames, Compat, DelimitedFiles
 
 set_default_plot_size(6inch, 3inch)
 
-plot(CSV.read(joinpath(dirname(@__DIR__), "data","issue120.csv")), x=:x1, y=:x2)
+dlm, varnames = readdlm(joinpath(dirname(@__DIR__), "data","issue120.csv"),
+                        ',', header=true)
+df = DataFrame(dlm, vec(Symbol.(varnames)))
+plot(df, x=:x1, y=:x2)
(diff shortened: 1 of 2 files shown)


test/testscripts/issue120.jl
```diff
@@ -1,5 +1,7 @@
-using Gadfly, DataFrames, Compat, CSV
+using Gadfly, DataFrames, Compat, DelimitedFiles
```
Mattriks: Why is there a `Compat ` here?

test/testscripts/issue120.jl
```diff
@@ -1,5 +1,7 @@
...
 
 set_default_plot_size(6inch, 3inch)
 
-plot(CSV.read(joinpath(dirname(@__DIR__), "data","issue120.csv")), x=:x1, y=:x2)
+dlm = readdlm(joinpath(dirname(@__DIR__), "data","issue120.csv"), ',')
```
Mattriks: or in this style (using `header=true`)
```julia
dlm, varnames =  readdlm( joinpath(...), ',', header=true)
```
which will simplify the next line (no convert needed for `dlm`)

Mattriks:
I just looked at #120, and it seems that the point was to test the syntax:
```julia
plot(readtable("test.csv"), x="x1", y="x2")
```
`readtable` produced a dataframe, so replacing with `DelimitedFiles.readdlm` (and constructing the dataframe outside the plot call) here isn't really the same.

test/testscripts/issue120.jl
Mattriks: Ignore my deleted remark on #120, I realise now #120 was a test about precision.

bjarthur:
i'm working on a second PR which removes all Compat.  no longer necessary now that we don't support 0.6.

PR #1356, opened 2019-11-28
hotfix/measure

CiaranOMara:
I've been using the new {h,v}band geometry for a day or so and have found that it errors in the presence of other geometries that use either xmin, xmax or ymin, ymax. This PR addresses that error.

# Contributor checklist:

# Proposed changes
- Define types of `Measure` as not concrete.

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -2,6 +2,7 @@
 isconcrete(x::T) where {T<:Number} = !ismissing(x) && isfinite(x)
 isconcrete(x::(Irrational)) = true
 isconcrete(x) = !ismissing(x)
+isconcrete(::Compose.Measures.Length) = false
 
 function isallconcrete(xs)
     ans = true
--- a/src/statistics.jl
+++ b/src/statistics.jl
@@ -824,10 +824,14 @@ function apply_statistic(stat::TickStatistic,
         vals = getfield(aes, var)
         if vals != nothing && eltype(vals) != Function && !(eltype(vals) <: Measure)
             if minval == nothing
-                minval = first(vals)
+                first_val = first(vals)
+                isa(first_val, Measure) && continue
+                minval = first_val
             end
             if maxval == nothing
-                maxval = first(vals)
+                first_val = first(vals)
+                isa(first_val, Measure) && continue
+                maxval = first_val
             end
             T = promote_type(typeof(minval), typeof(maxval))
             T = promote_type(T, eltype(vals))
--- a/test/testscripts/hvband.jl
+++ b/test/testscripts/hvband.jl
@@ -5,4 +5,9 @@ p1 = plot(xmin=[1.0, 5.0, 7.0], xmax=[2.0, 6.5, 8.0] , Geom.vband, Theme(default
 
 p2 = plot(ymin=[2.5], ymax=[7.5], Geom.hband, Theme(default_color="red"));
 
-hstack(p1, p2)
+p3 = plot(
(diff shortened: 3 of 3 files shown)


Mattriks:
Can you give example code of the sort of errors you get (preferable open a new issue).  `Measure` types (from `Measures.jl`) should just work (similar to units in `Unitful.jl`), and it would be better to add the needed operations to `Measures.jl` such as:
```
 Base.oneunit(a::Type{Length{U,T}}) where {U,T} = Length(U, one(T))
 Base.isfinite(a::Length) = isfinite(a.value) 
```

CiaranOMara:
If I understand you correctly, you are presenting the prospect of aesthetics taking on values such as cm and possibly other units such as degrees Celsius, which would be really nice.

```julia
using Compose, Gadfly, Unitful
import Unitful:°C
plot(x=[1cm,2cm], y=[1°C,2°C], Geom.point)
```

In which case, I think other design questions to address while forming an outline are:
- how to identify and group incompatible types?
- how should incompatible types scale on a shared axis?
- how should a shared axis be presented?

The issue that I responded to here was that type `Number` does not convert to type `Measure` but needs to share an axis. Additionally, the only values of `Measure` I had considered were 0w, 1w, 0h, and 1h in the context of the `BandStatistic`. In the delimited case of `BandStatistic `, I think skipping over type `Measure` in `Stat.TickStatistic` works as a hotfix.

Mattriks:
The purpose of `apply_statistic(TickStatistic)`, is to define the range and tickmarks of the x and y axes. Consider these two (separate) examples:

1) One layer has `xmin,ymin` as a Float and another layer has `xmin,ymin` as a relative Measure type (e.g. 1h or 1w as in your `Geom.band` example). In this case, the relative Measure doesn't provide info about the absolute range of `x,y`.

2) `plot(x=1:5, y=rand(5), size=[11:2:20;].*mm, Geom.point)`  In this case, the absolute Measure (in the size aesthetic) is relevant for choosing the range of the `x,y` axes.

To make `Stat.TickStatistic` compatible with both these examples, it seems we need to handle the relative and absolute Measure types differently in `apply_statistic(TickStatistic)`.

CiaranOMara:
So in example 1, the absolute and relative measures were incompatible, and there was no need to provide ticks or labels for the relative measures. It is, however, conceivable to use relative measures and want ticks with percentage or fraction labels.

In example 2, there is a requirement to keep track of the leftmost, rightmost, topmost, and bottommost outer geometry. But things like highlights or padding may also modify these geometries.

Also, If we go down the path of including other units, there is a scenario where there are units that should scale relative to other units. For example, imperial and metric units — should they ever be mixed into the same plot. Indeed, a user may want to scale an axis of their own accord.

I agree that incompatible types/units (Number, AbsoluteMeasure, etc.) cannot be mixed; they somehow need to be separated, then handled. I think the problem then becomes a question of how to present the ticks of different units on a shared axis.

Mattriks:
By "relative Measure types" I mean in `Measures.jl` (see my proposed [new solution](https://github.com/GiovineItalia/Gadfly.jl/issues/1357#issuecomment-559635064)). 
Gadfly is compatiible e.g. with `Unitful` percent (see [discourse post](https://discourse.julialang.org/t/scale-y-continuous-percentage-format/31397/5)).

CiaranOMara:
Closing in favour of https://github.com/GiovineItalia/Gadfly.jl/pull/1358.

PR #1364, opened 2019-12-16
add TagBot

bjarthur:


bjarthur:
a second commit changed CompatHelper frequency from every minute (doh!) to once per week.

bjarthur:
a third commit changes CI to test on julia 1.3

.github/workflows/CompatHelper.yml
```diff
@@ -2,7 +2,7 @@ name: CompatHelper
 
 on:
   schedule:
-    - cron: '* * * * 0'
+    - cron: '0 0 * * 0'
```
tlnagy: Isn't this once per day?

bjarthur: midnight on sunday, so once per week.  https://crontab.guru/#0_0_*_*_0

bjarthur: it's a shame that both tagbot and compathelper poll.  would be nice if they could be re-engineered to use interrupts.

bjarthur:
worth noting that the change to travis specifies "julia 1", not "julia 1.3".  so when 1.4 etc is released it will automatically start testing on it.  just learned this was possible from tim holy's work on indirectarrays.jl.

Issue #1366, opened 2019-12-18
making scatter plots with different shaped markers (issues)


SimonEnsemble:
# issue 1

```
plot(x=rand(4), y=rand(4), Geom.point, shape=[Shape.hexagon],
     Guide.xlabel("Re"),
     Guide.ylabel("Im")
)
```
automatically includes a legend, which I don't want. this seems an odd default, and how can I remove the legend?

![image](https://user-images.githubusercontent.com/30643894/71052076-323f9b00-20ff-11ea-88aa-51ce8cd8c12a.png)

# issue 2
this code results in no output
```
plot(layer(x=rand(4), y=rand(4), Geom.point, shape=[Shape.hexagon]),
     layer(x=rand(4), y=rand(4), Geom.point, shape=[Shape.cross]),
     Guide.xlabel("Re"),
     Guide.ylabel("Im")
)
```
how can I make two layers with different symbols outside (without building a `DataFrame`)?

# issue 3
the markers in the key seem to be a different size than inside the plot. is it possible to make them the same size using `Theme.point_size`?

![image](https://user-images.githubusercontent.com/30643894/71052313-d3c6ec80-20ff-11ea-98f2-2caee6d8c618.png)

# issue 4
even in the `DataFrame` paradigm, 

```
df = DataFrame(x=rand(3), y=rand(3), z=["a", "a", "b"])

plot(df, x=:x, y=:y, shape=:z, hipster_theme, shapes=[Shape.cross, Shape.circle])
```
how can I control which shapes are used, e.g. specify crosses and circles?

![image](https://user-images.githubusercontent.com/30643894/71052524-6ebfc680-2100-11ea-8a6a-7551144857ea.png)

# issue 5
is there a manual key (legend) I could make while specifying the shapes of the data? That way I could get rid of the title on the legend above, which is "z".

Mattriks:
1) and 4.  use `Theme()` e.g. 
`Theme(key_position=:none, point_shapes=[Shape.cross, Shape.circle])`
To see all the options do `dump(Theme(), maxdepth=1)`

2) Has been fixed recently on master branch (#1333). To use master branch do
```
]add Gadfly#master
```

I'll get back to you about 3 and 5.

Mattriks:
[...]

PR #1370, opened 2019-12-21
more Discrete Scale documentation

Mattriks:
- [x] I've updated the documentation

### This PR:
- adds documentation pertaining to #1366
- the main remaining point in #1366 is "issue 3" which I leave  to a future PR (size revamp).

### DIFF
--- a/docs/src/gallery/scales.md
+++ b/docs/src/gallery/scales.md
@@ -162,6 +162,26 @@ p2 = plot(dataset("datasets", "CO2"), x=:Conc, y=:Uptake,
 hstack(p1,p2)
 ```
 
+## [`Scale.shape_discrete`](@ref)
+
+```@example
+using Gadfly, RDatasets, Statistics
+set_default_plot_size(14cm, 8cm)
+
+tips = dataset("reshape2", "tips")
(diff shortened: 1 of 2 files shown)


Mattriks:
Gadfly's Compose compat obviously needs to be bumped,. Would it also make things easier if common packages (e.g. Colors etc)  have the same compats?

bjarthur:
herein lies the problem with the polling (instead of interrupt style) mechanism used by CompatHelper.  ideally the tagging of Compose should have triggered an immediate notification to Gadfly that it's compat section is out of date.  but Github permissions don't allow actions to post PRs to other repos.  so all CompatHelper do is poll periodically.  Gadfly and Compose are currently set to do so midnight on sunday, or whenever an issue is opened or re-opened.

bjarthur:
just add another commit which updates Compose in compat.

tlnagy:
> Gadfly and Compose are currently set to do so midnight on sunday, or whenever an issue is opened or re-opened.

Why can't we just poll daily? It would be nice to batch the updates more than that, but it's not really a hassle.

bjarthur:
we could, i just figured it's a load on the servers that isn't warranted, despite being free.

Mattriks:
Rebased and ready to merge.

Issue #859, opened 2016-06-30
PDF with Cairo Overlapping Legend with :top position.


Nosferican:
Using Julia 4.6 with Cairo in order to save a plot to disk as a pdf. When the legend is on the right there is no problem, but if placed on top, the legend symbols and text overlap within columns.

[normalcdf.pdf](https://github.com/dcjones/Gadfly.jl/files/340472/normalcdf.pdf)

mortenpi:
Could this be a dup of #688? In that instance the solutions was to make sure that `Fontconfig` is installed and the precompilation cache updated.

Mattriks:
There is more to this issue than just `Fontconfig` (see also discussion in https://github.com/GiovineItalia/Gadfly.jl/issues/688#issuecomment-324192928).
In [this line](https://github.com/GiovineItalia/Gadfly.jl/blob/bf50828f8603ab0b4f918b5247d9c99186960876/src/guide.jl#L220) the sum should be `entry_height + xpad`.  The key column width should include (a hypothetical square around the swatch) + (text label width). The calculated column width is too short, hence the text overlaps with the next column.

PR #1379, opened 2020-01-05
Size revamp 2: Guide.sizekey

Mattriks:

### This PR
- The second in a set of 3 PRs. This PR adds `Guide.sizekey` 
- is another step towards closing issues #766, #1366

This PR makes changes to the layout of shapekey and sizekey, via `render_discrete_key`. This doesn't affect colorkey, which currently uses `render_discrete_color_key`. `render_discrete_color_key` contains bugs (#859), that become worse with larger swatch sizes.  The long-term plan is that all discrete aesthetics (including `color`) will use `render_discrete_key`.

### Examples
```julia
labels = rand(randstring.(fill(8, 5)), 100) 
datas, aes = Gadfly.Data(size=labels), Gadfly.Aesthetics()
sc1 = Scale.size_discrete2(n->range(4pt, 8pt, length=n))
Scale.apply_scale(sc1, [aes], datas)
 gsk = Guide.sizekey(title="Size key")
theme = Theme(key_swatch_shape=Shape.hexagon, key_label_font_size=10pt, 
    discrete_highlight_color=identity, alphas=[0.1], key_max_columns=5)
ctxs =  Guide.render(gsk, theme, aes)[1].ctxs
vstack(hstack(ctxs[1:2]...), ctxs[5]) 
```
sizekey

Example from `gallery/guides.md` (in this PR):
```julia
plot(Titanic, Scale.x_log10,  Scale.y_log10,
    x=:Freq, y=:prcnt, color=:Age, shape=:Sex,  size=:Class,
    Scale.size_discrete2(sizemap), Guide.sizekey(title="Passenger\n Class"),
    Guide.colorkey(pos=[0.1, -0.3h]), Guide.shapekey(pos=[0.5, -0.31h]),
    Guide.ylabel("% of Passenger Class"),
 Theme(discrete_highlight_color=identity, alphas=[0.1], key_swatch_color="grey",
    key_swatch_shape=Shape.circle, point_size=3pt) )
```
Titanic

### DIFF
--- a/docs/src/gallery/guides.md
+++ b/docs/src/gallery/guides.md
@@ -67,6 +67,28 @@ plot(Dsleep, x=:BodyWt, y=:BrainWt, Geom.point, color=:Vore, shape=:SleepTime,
 ```
 
 
+## [`Guide.sizekey`](@ref)
+
+```@example
+using Compose, Gadfly, RDatasets
+set_default_plot_size(14cm, 8cm)
+
+Titanic = dataset("datasets", "Titanic")
+Class =  by(Titanic, :Class, :Freq=>sum)
+Titanic = join(Titanic[Titanic.Survived.=="Yes",:], Class, on=:Class)
+Titanic.prcnt = 100*Titanic.Freq./Titanic.Freq_sum
+sizemap = n->range(3pt, 8pt, length=n)
+
+plot(Titanic, Scale.x_log10,  Scale.y_log10,
+    x=:Freq, y=:prcnt, color=:Age, shape=:Sex,  size=:Class,
+    Scale.size_discrete2(sizemap), Guide.sizekey(title="Passenger\n Class"),
+    Guide.colorkey(pos=[0.1, -0.3h]), Guide.shapekey(pos=[0.5, -0.31h]),
+    Guide.ylabel("% of Passenger Class"),
+ Theme(discrete_highlight_color=identity, alphas=[0.1], key_swatch_color="grey",
+    key_swatch_shape=Shape.circle, point_size=3pt) )
+```
+
+
 ## [`Guide.title`](@ref)
 
 ```@example
--- a/src/guide/keys.jl
+++ b/src/guide/keys.jl
@@ -72,7 +72,7 @@ function render_discrete_key(labels::Vector{String}, title_ctx::Context, title_w
     shape1 = shapes[1]
     shapes = (shape1==nothing) ? fill(theme.key_swatch_shape, n) : theme.point_shapes[shapes]
     (colors[1]==nothing) && (colors = fill((theme.key_swatch_color==nothing) ? theme.default_color : theme.key_swatch_color, n))
-    (sizes[1]==nothing) && (sizes = fill(theme.point_size, n))
+    (sizes[1]==nothing) && (sizes = fill((theme.key_swatch_size==nothing) ? theme.point_size : theme.key_swatch_size, n))
 
     # only consider layouts with a reasonable number of columns
     maxcols = theme.key_max_columns < 1 ? 1 : theme.key_max_columns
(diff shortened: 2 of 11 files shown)


Mattriks:
### To do:
- [ ] change the key swatch stroke width (to `theme.highlight_width`)

Mattriks:
Placeholder

docs/src/gallery/guides.md
```diff
@@ -67,6 +67,28 @@ plot(Dsleep, x=:BodyWt, y=:BrainWt, Geom.point, color=:Vore, shape=:SleepTime,
...
+Titanic.prcnt = 100*Titanic.Freq./Titanic.Freq_sum
+sizemap = n->range(3pt, 8pt, length=n)
+
+plot(Titanic, Scale.x_log10,  Scale.y_log10,
+    x=:Freq, y=:prcnt, color=:Age, shape=:Sex,  size=:Class,
+    Scale.size_discrete2(sizemap), Guide.sizekey(title="Passenger\n Class"),
+    Guide.colorkey(pos=[0.1, -0.3h]), Guide.shapekey(pos=[0.5, -0.31h]),
+    Guide.ylabel("% of Passenger Class"),
+ Theme(discrete_highlight_color=identity, alphas=[0.1], key_swatch_color="grey",
+    key_swatch_shape=Shape.circle, point_size=3pt) )
```
bjarthur: a near identical example appears in gallery/scales, albeit without the key.  is there another dataset that could be used here?  the more variety in the gallery the more likely the reader might find something closer to what they're looking for.

src/guide/keys.jl
```diff
@@ -40,7 +40,7 @@ function Guide.render(guide::Guide.ShapeKey, theme::Gadfly.Theme, aes::Gadfly.Ae
         colors = collect(keys(aes.color_key_colors))
     end
     
-    title_context, title_width = Guide.render_key_title(guide_title, theme)
```
bjarthur: is `render_key_title` to be replaced by `render_key_title2` in gadfly 2.0?  if so, might add a comment in the code about that somewhere.  and create an issue to which we can add the 2.0 milestone tag.

Mattriks: `render_key_title2` goes with `render_discrete_key`, whereas `render_key_title` is used by  `render(guide::ColorKey, ...)`.  I plan to make all the _discrete_ asesthetics use `render_discrete_key`, after which only the _continuous_ color key will be using the current `render_key_title`.  I don't have any plans for changing the continuous color key at the moment.

docs/src/gallery/guides.md
Mattriks: With this PR, the size key appears* in the example in gallery/scales, but there are a few differences including:
- key title is different (Class changed to "Passenger\nClass" in gallery/guides)
- the swatch shape is different (squares is the default)
- `levels` is set in `Scale.size_discrete` in gallery/scales

Thus the differences between the two examples in gallery/scales and gallery/guides illustrates to users the difference between a Scale and a Guide.

*Note that the size key is triggered in the same way as the color keys, i.e the size key is **not** triggered by one element vectors e.g. `size=[1.41]`

Mattriks:
The only q remaining for me in this PR is: can the default point size be changed from 0.95mm to 3pt (~1mm).?  The reason is that with this PR the key swatch size defaults to `Theme(point_size)`, and 3pt is closer to the current key swatch size (0.95mm is slightly smaller than the current key swatch size).

src/guide/keys.jl
```diff
@@ -166,16 +164,96 @@ function render_discrete_key(labels::Vector{String}, title_ctx::Context, title_w
...
+        title_form = text(0.0w, title_height, title, hleft, vbottom)
+    elseif theme.guide_title_position == :center
+        title_form = text(0.5w, title_height, title, hcenter, vbottom)
+    elseif theme.guide_title_position == :right
+        title_form = text(1.0w, title_height, title, hright, vbottom)
+    else
+        error("$(theme.guide_title_position) is not a valid guide title position")
+    end
+
+    title_padding = 1.5mm
```
bjarthur: the only difference i can see between `render_key_title` and `render_key_title2` is the use of `4mm` in the former here.  to eliminate such code duplication, would it be possible to pass in `title_padding` as an argument?

Mattriks: I would prefer to keep the two functions separate for the moment, as I'm working towards separating the keys for discrete and continuous aesthetics.  After I complete that I'm happy to tidy up whatever's remaining.

Mattriks:
Please answer my question above about the default point size.

Issue #1368, opened 2019-12-19
UndefVarError: @colorant_str not defined


etorkia:
I have a series of charting functions built using Gadfly... Suddenly I am having an issue with the Scale.color_discrete_manual function that is returning : UndefVarError: @colorant_str not defined. 

There is also an issue with the recognition of **colorant"deep sky blue"** as an available color. This error is preventing the entire package from precompiling. 

Here is the offending plot:
```julia
plot(graph_tbl, y=:name, x=:correlation, Guide.Title("Variables with Biggest Impact"),
Guide.xlabel("Rank Correlation"), Guide.ylabel("Input"), Geom.bar(orientation=:horizontal),
Theme(bar_spacing=10pt), color=:impact,
Scale.color_discrete_manual(colorant"red", colorant"deep sky blue"),
Scale.x_continuous(minvalue=-1, maxvalue=1))
```
Here is the stack trace:

```
ERROR: LoadError: LoadError: LoadError: UndefVarError: @colorant_str not defined
Stacktrace:
 [1] top-level scope
 [2] #macroexpand#32 at .\expr.jl:107 [inlined]
 [3] macroexpand at .\expr.jl:106 [inlined]
 [4] docm(::LineNumberNode, ::Module, ::Any, ::Any, ::Bool) at .\docs\Docs.jl:509 (repeats 2 times)
 [5] @doc(::LineNumberNode, ::Module, ::String, ::Vararg{Any,N} where N) at .\boot.jl:461
```

Any help would be great!

bjarthur:
possibly related to [Colors 0.10 failing](https://github.com/GiovineItalia/Gadfly.jl/pull/1367) and [Colors 0.11 working](https://github.com/GiovineItalia/Gadfly.jl/pull/1369).  see this [Colors.jl issue](https://github.com/JuliaGraphics/Colors.jl/pull/383).

kimikage:
I'm sorry but I *breaked* the color parser (https://github.com/JuliaGraphics/Colors.jl/pull/360)
"deep sky blue" is invalid on Colors.jl v0.10 and *v0.11*. Should I fix this problem by reverting?

Mattriks:
x11 color names are user-friendly, I think that most users are not e.g. RGBA aficionados.

kimikage:
I agree with you.
[...]

PR #1383, opened 2020-01-11
update compat to Colors v0.10

bjarthur:
tests pass locally for me.  @kimikage 

does this fix https://github.com/GiovineItalia/Gadfly.jl/issues/1368 ?


kimikage:
I guess that the cause of `UndefVarError: @colorant_str not defined` is a failure of package compilation. Since I don't fully understand the cause of the compilation error (i.e. since reproduction codes are not shown), I cannot say for certain.
However, this PR definitely improves the situation. *If* this doesn't solve the issue #1368 then there should be something else.

PR #1384, opened 2020-01-11
fix DataFrames deprecation warnings

bjarthur:
`Not` was introduced in DataFrames 0.19, so all prior versions are incompatible.

should probably wait to merge this until after the Colors issue as been resolved and a release tagged with it.


bjarthur:
found a way to maintain DataFrame compatibility back to 0.15, so can include this PR in the next release

Issue #380, opened 2014-07-30
Horizontal box plots?


sjl124:
HI there,

As far as I can tell (I am a newbie though) there's no way to make horizontal box plots. Is that correct? Boxplot does not seem to take an 'orientation' argument, nor is there the coord_flip capability of ggplot.

Would be great if this feature could be added! :)

Steve

PR #1388, opened 2020-01-21
resolve issue #1374: document color spec.

### LINKED ISSUE #1374: document color specification
it would be helpful if the docs at http://gadflyjl.org listed the allowed color values e.g. as a table, so that it would be evident whether a usage is valid.

e.g. is there a reference i could use to check whether `Theme( default_color="foo" )` or `Scale.lab_gradient("foo", "bar", "baz")` produces a valid style element?

i saw that a related issue (#254) had been opened and closed back in 2014. i also noticed this mention in the tutorial:

> Color scales in Gadfly by default are produced from perceptually uniform colorspaces (LUV/LCHuv or LAB/LCHab), though it supports RGB, HSV, HLS, XYZ, and converts arbitrarily between these. Of course, CSS/X11 named colors work too: "old lace", anyone?

which gives a hint of where i might look -- but it felt like a circuitous path to take just to find out whether, for example, `"cyan"` or `"darkred"` were recognizable values.

ben-schulz:
Fixes #1374

# Contributor checklist:

# Proposed changes

and 

- add hyperlink to table in `Colors.jl`,
- note the `Colorsjl` library as the point-of-truth.
- incorporate revision suggested by @Mattriks.

### DIFF
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -159,8 +159,8 @@ Ah, a scientific discovery: Setosa has short but wide sepals!
 
 Color scales in Gadfly by default are produced from perceptually uniform
 colorspaces (LUV/LCHuv or LAB/LCHab), though it supports RGB, HSV, HLS, XYZ, and
-converts arbitrarily between these. Of course, CSS/X11 named colors work too:
-"old lace", anyone?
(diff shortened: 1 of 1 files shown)


ben-schulz:
also addresses #380

bjarthur:
thanks for this!

also, in general it is preferred to make a change to an existing PR by using `git rebase` and `git push --force` than to close the PR and open a new one.  not that big of a deal, but it's easier to keep track of comments that way.

ben-schulz:
my mistake; i'll follow that procedure next time. thanks!

PR #1397, opened 2020-02-15
get docs to auto build again

bjarthur:
per https://github.com/marketplace/actions/julia-tagbot#personal-access-tokens


bjarthur:
i've also manually added a git tag for 1.2 since tagbot inexplicably failed.

PR #1408, opened 2020-03-28
Theme grid order and panel line width

### LINKED ISSUE #1407: Possible improvements for panel border
When I added a border to the panel of the plot using `panel_stroke=colorant"black"`. I noticed that the lines and the grids are on the top of the borders, which seems not very pretty. Is this intended or will it be fixed in the future? Also, I wonder if you can add another function to adjust the thickness of the border. Thank you!

Mattriks:

### This PR:
- closes #1407
- adds `Theme(grid_line_order=0)`
- adds `Theme(panel_line_width=0.3mm)`
- docstrings added to Theme

### Example:

In this example, the grid lines are behind the panel frame.
```julia
plot(x=[5100, 5400], y=[0, 1], Geom.blank,
    Guide.yticks(ticks=0:0.1:1),
    Theme(grid_line_style=:solid, grid_line_width=0.5mm,
        grid_color="gray92", panel_stroke="black", 
        grid_line_order=-2,  panel_line_width=1mm)
)
```
gridstyle

### DIFF
--- a/src/guide.jl
+++ b/src/guide.jl
@@ -147,6 +147,7 @@ function render(guide::PanelBackground, theme::Gadfly.Theme,
                     svgclass("guide background"),
                     stroke(theme.panel_stroke),
                     fill(theme.panel_fill),
+                    linewidth(theme.panel_line_width),
                     svgattribute("pointer-events", "visible"))
 
     return [PositionedGuide([back], 0, under_guide_position)]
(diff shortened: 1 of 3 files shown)


Mattriks:
Tests are passing now with RDatasets 0.6.7.  For this PR, I just need to check off the regression tests.

Mattriks:
The regression test pass.  The only other testscript to use `panel_stroke` is "subplot_layers.jl".  For "subplot_layers.jl" there is a difference in the panel border (slightly thinner for devel-output compared to master-output - for master the panel border must be inheriting it's linewidth from somewhere, because it's not explicitly set). But I've tested this PR with `Geom.subplot_grid`, and the new features `panel_line_width` and `grid_line_order` behave as expected.  So this PR is ready to merge.

Mattriks:
bump @bjarthur

Issue #1385, opened 2020-01-18
remove code duplication between render_key_title and render_key_title2


bjarthur:
per https://github.com/GiovineItalia/Gadfly.jl/pull/1379#discussion_r368237119

Mattriks:
Following #1386 here is a list of things to be done (in no order):

- [ ] take absolute point sizes into account in setting axis limits
- [ ] Syntax `Guide.colorkey(nothing)` to remove individual keys
- [ ] fix triggering of `Guide.shapekey` (#1366)
- [ ] Switch discrete `color` aesthetic to use `render_discrete_key` (will fix #859)
- [ ] key stacking (#766)
- [ ] tidy up render_key_title and render_key_title2

And in the distance:
- [ ] Allow 1 discrete color key and 1 continuous color key per plot (#1098, #1171)

PR #1411, opened 2020-04-06
Guide.shapekey updates

### LINKED ISSUE #1366: making scatter plots with different shaped markers (issues)

Mattriks:

### This PR:
- closes #1366 (see #1385 Roadmap)
- fixes triggering of shapekey (issue 1 in #1366, plot p1 below)
- fixes issue 6 in #1366 (plot p2)
- ~~shapekey integration with 1-length `size` and `color` vectors (plot p3)~~

### Example

```julia
x, y = 1:5, rand(5)
p1 = plot(x=x, y=y, shape=[Shape.square], size=[8pt])

p2 = plot(Theme(point_size=6pt, alphas=[0.6]),
    layer(x=0.5:4, y=rand(4), shape=["Gr1"]),
     layer(x=4*rand(4), y=rand(4), shape=["Gr2"]) )

p3 = plot(x=x, y=y, shape=x, size=[6pt], color=[colorant"orange"])
hstack(p1,p2,p3)
```
key_shape

### DIFF
--- a/src/misc.jl
+++ b/src/misc.jl
@@ -338,6 +338,6 @@ function discretize_make_ia(values::CategoricalArray{T}, levels::Vector) where {
     return IndirectArray(index, convert(Vector{T},levels))
 end
 function discretize_make_ia(values::CategoricalArray{T}, levels::CategoricalVector{T}) where T
-    _levels = map!(t -> ismissing(t) ? t : get(t), Vector{T}(length(levels)), levels)
+    _levels = map!(t -> ismissing(t) ? t : get(t), Vector{T}(undef, length(levels)), levels)
     discretize_make_ia(values, _levels)
 end
--- a/docs/src/gallery/shapes.md
+++ b/docs/src/gallery/shapes.md
@@ -1,11 +1,20 @@
 # Shapes
 
-## [`Shape.square`](@ref)
+## [`Shape.square`](@ref) and friends
 
 ```@example
 using Gadfly, RDatasets
-set_default_plot_size(14cm, 8cm)
-plot(dataset("HistData","DrinksWages"),
-     x="Wage", y="Drinks", shape=[Shape.square],
-     Geom.point, Scale.y_log10)
+set_default_plot_size(21cm, 8cm)
+
+p1 = plot(dataset("HistData","DrinksWages"), x="Wage", y="Drinks", 
+    shape=[Shape.square], Scale.y_log10)
+
+aww, mws = dataset("MASS", "Animals"), dataset("quantreg", "Mammals")
+p2 = plot( layer(aww, x=:Body, y=:Brain, shape=["Brain weight"]),
+    layer(mws, x=:Weight, y=:Speed, shape=["Run speed"]),
+    Scale.x_log10, Scale.y_log10, Guide.xlabel("Body weight"),
+    Guide.ylabel("Brain weight and Run speed"),
+    Theme(point_shapes=[Shape.circle, Shape.star1], alphas=[0.0],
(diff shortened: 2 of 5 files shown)


Mattriks:
The regression tests showed the following changes. With this PR, in testscripts:
1. "point_color_shape_size.jl", "point_shape_explicit.jl", and "shape_hvline.jl" the shapekey is not triggered (this is correct). Note that the examples in the first two testscripts are written in "old Gadfly style" (pre-shapekey) and could be updated in the future.

2. "key_columns.jl"  shows that the size aesthetic will override any default settings in `Theme`.  This maybe an issue in a plot that uses multiple aesthetics (e.g. distinct size and shape mappings).  Need to think more about plot p3 above (and if necessary add more to this PR).

tlnagy:
This is great! I'm very excited for the propagation of a single color and shape through a layer. Makes multiple layer plots a lot easier. How hard would it be to autowrap non-array values? So you could do something like `color=colorant"orange"` instead of `color=[colorant"orange"]`, which makes more sense from a user's perspective.

Mattriks:
placeholder

Mattriks:
The regression tests now show changes to just the 3 testscripts listed in point 1 in my regression test post above. With this PR, in those 3 testscripts the shapekey doesn't trigger with `shape=[Shape.function]`. This is consistent with e.g.`color=[colorant"color"]` which doesn't trigger the colorkey.  

This PR is now ready to merge.

src/misc.jl
```diff
@@ -338,6 +338,6 @@ function discretize_make_ia(values::CategoricalArray{T}, levels::Vector) where {
     return IndirectArray(index, convert(Vector{T},levels))
 end
 function discretize_make_ia(values::CategoricalArray{T}, levels::CategoricalVector{T}) where T
-    _levels = map!(t -> ismissing(t) ? t : get(t), Vector{T}(length(levels)), levels)
+    _levels = map!(t -> ismissing(t) ? t : get(t), Vector{T}(undef, length(levels)), levels)
```
bjarthur, line 341: good catch!  and great to see that codevec shows this line is now tested.

bjarthur:
is it worth putting this in the gallery?

Mattriks:
Done.  On Travis, for some reason the Codecov bit failed on Julia 1 now.

Mattriks:
This is ready to merge.

PR #1416, opened 2020-04-13
bump compat for Unitful to 1.0

Mattriks:

- I also added `set_default_plot_size` to the 3 Unitful testscripts, so that they produce clearer test plots.


bjarthur:
strange that compathelper didn't do this automatically.

Mattriks:
@bjarthur Unitful isn't in [deps], and doesn't need to be.

PR #1424, opened 2020-04-24
Enable subplot axis label orientation

### LINKED ISSUE #424: Axis label orientation in subplots
Is there a way to set the orientation for the actual axis labels in subplots? I find that `Guide.xlabel()` and `Guide.ylabel()` apply to the overall label for all subplots and not the individual subplots' axis labels.

Thanks!

Mattriks:

### This PR:
- enables changing the axis label orientation in subplots 
- fixes #424
- adds documentation about subplot inner and outer guides.

### Example
```julia
using RDatasets
haireye = dataset("datasets","HairEyeColor")
palette = ["brown","blue","tan","green"]

plot(haireye, y=:Sex, x=:Freq, color=:Eye, ygroup=:Hair,
    Geom.subplot_grid(Geom.bar(position=:stack, orientation=:horizontal),
        Guide.ylabel(orientation=:vertical) ),
    Scale.color_discrete_manual(palette...),
    Guide.colorkey(title="Eye\ncolor"),
    Guide.ylabel("Hair color"), Guide.xlabel("Frequency")
)
```
haireye

### DIFF
--- a/docs/src/gallery/geometries.md
+++ b/docs/src/gallery/geometries.md
@@ -487,7 +487,7 @@ plot(x=rand(25), y=rand(25), Geom.step)
 ```
 
 
-## [`Geom.subplot_grid`](@ref)
+## [[`Geom.subplot_grid`](@ref)](@id Gallery_Geom.subplot_grid)
 
 ```@example
 using Gadfly, RDatasets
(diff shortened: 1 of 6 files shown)


bjarthur:
LGTM.  my only comment is that it would be nice to have the example in the docs be different than that in the tests.  tests can double as docs, but only to the extent that they are different.

Mattriks:
I thought the example was too nice  😃  I might add revamping `subplot_grid` to the next Roadmap (after #1385), to continue what I've started here, in which case I'll change/update the testscripts then.

Issue #601, opened 2015-06-03
Cursor over Gadfly plot momentarily freezes Juno


dlfivefifty:
Whenever I have a Gadfly plot, Juno will continually pause as it tries to switch to "zoom" mode on the Gadfly plot. This is in OS X.  I'm not sure whether this is an issue with Gadfly or with Juno.  

See also  

https://github.com/one-more-minute/Julia-LT/issues/193

dcjones:
This is happening in ipython too, so I think it's Gadfly's fault, and I think I know why. I'll fix it when I have a chance later today.

dlfivefifty:
Thank you!!!

dlfivefifty:
This issue is actually still there in Juno when the plot is complicated.  Can we reopen it?

dcjones:
Yeah, I'll have another look.

dlfivefifty:
Is it possible to disable interactive graphs for the moment until this is fixed?

dcjones:
Yep, just call `set_default_plot_format(:svg)` before plotting.

dlfivefifty:
Thanks!  that stopped it from freezing

PR #1445, opened 2020-05-28
Add disable keymap settings

TsuMakoto:
# Contributor checklist:

# Situation
First, enter characters in the cell with the jupyter notebook.
At that time, if you enter the assigned key (-, +, i, o etc ...), the event of javascript will be triggered and unexpected behavior will occur.
There will also be a delay.
Gadfly

This is one example.
But if it is rendered multiple times, it will be further delayed.

# Proposed changes
- Defined a function that disable a keymap.
- Gadfly.plotroot_keymap(false): disable keymap

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@ Each release typically has a number of minor bug fixes beyond what is listed her
 
 # Version 1.x
 
+ * Add `plotroot_keymap(enable::Bool)` (#1445)
  * Add `Guide.manual_discrete_key`, updates `Guide.manual_color_key` (#1441)
  * Increase support for syntax `color=[colorant"color"]` (#1438)
  * Enable `color` aesthetic for `Stat.qq` (#1434)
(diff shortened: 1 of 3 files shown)


Mattriks:
In Jupyter, why not just do:
```julia
p = plot(...)
draw(SVG(), p)
```
to turn off SVGJS functionality for a plot?

TsuMakoto:
@Mattriks 
Thanks for your suggestions.

[IJulia](https://github.com/GiovineItalia/Gadfly.jl/blob/fcf7596cdbb3eff5a4302067a76eb7e966d0c24b/docs/src/man/backends.md#ijulia)

You mean this document, right?

I feel a hassle that call the draw function every time.
I want to add a setting that disables all at once.

Is there any good way?

Mattriks:
A switch that allows you choose between PDF, PNG, SVG, would be a better way to do this.  There is an old command  `set_default_plot_format()`  which use to work (e.g. https://github.com/GiovineItalia/Gadfly.jl/issues/601#issuecomment-136478918), but it's functionality was lost over time (see [Julia's display model](https://docs.julialang.org/en/v1/base/io-network/#Multimedia-I/O-1)). I showed one way of making it work in the REPL: https://discourse.julialang.org/t/can-i-choose-display-software/29510.

Someone would have to investigate further about how to make it work in Jupyter/IJulia, and I think that would be a better way to go than this PR.

TsuMakoto:
OK
I hope that the ones supported by jupyter / IJulia will be implemented.
Until then, it will be handled with `draw(SVG(), p)`.

Thank you.

PR #1447, opened 2020-05-31
Bump compat for CategoricalArrays to 0.8

Mattriks:

### This PR
- bumps compat for `CategoricalArrays` to 0.8
- and `DataFrames` to 0.21


Mattriks:
Once this is merged, we should do a new release in about a week or so.

bjarthur:
or...  wouldn't it have been better to tag a release *before* this was merged so that people who do not want to upgrade CategoricalArrays and DataFrames can get the latest Gadfly features and bug fixes.  is it possible to tag a release of a commit that is not master?

Mattriks:
The compats have to be bumped sometime!  My thoughts were that the imminent minor release (1.3) should include this compat bump to CategoricalArrays 0.8.  Also the next minor release (1.4)  will probably include DataFrames 1.0.

bjarthur:
could have had it both ways by tagging 1.3 before this merge, and then 1.3.1 immediately after.  dependency compatibility is a huge pain sometimes so it's important to give people both options.

Issue #818, opened 2016-03-31
Is there an easy way to pass aesthetics?


idmit:
Hi! I have code like this:
```julia
using Gadfly
using RDatasets

iris = dataset("datasets", "iris")

plot(iris, x = :SepalLength, y = :PetalLength, Geom.point, Geom.smooth)
```

I'd like to color points with blue and line with red. I know that this code does what I want:
```julia
plot(iris, 
layer(x = :SepalLength, y = :PetalLength, Geom.point, Theme(default_color = colorant"blue")), 
layer(x = :SepalLength, y = :PetalLength, Geom.smooth, Theme(default_color = colorant"red"))
)
```

But here is a list of things I tried intuitively before coming to this.
1.
```julia
plot(iris, x = :SepalLength, y = :PetalLength, Geom.point, color = colorant"blue", 
Geom.smooth, color = colorant"red")
```
2.
```julia
plot(iris, x = :SepalLength, y = :PetalLength, Geom.point(aes(color = colorant"blue")), 
Geom.smooth(aes(color = colorant"red")))
```
3.
```julia
plot(iris, x = :SepalLength, y = :PetalLength,
layer(Geom.point, color = colorant"blue"), 
layer(Geom.smooth, color = colorant"red")
)
```

I think that the second one feels like the most reasonable approach. It strongly resembles ggplot2 way of doing things. Does Gadfly have DataType for Aesthetics?

tlnagy:
I agree that the `Theme(default_color=colorant"blue")` isn't very intuitive, personally I think the 3rd option makes the most sense and feels the most "Gadfly"esque.

tlnagy:
I'm interested in implementing this since this makes Gadfly more intuitive and easier to pick up.

shashi:
I agree that `layer(x = :SepalLength, y = :PetalLength, Geom.point, Theme(default_color = colorant"blue"))` is verbose for each different thing you want to plot. The awkwardness seems to come from the fact that you have to specify `x` and `y` again.

It looks like there is a pattern we can build around option 3. However, I don't know how to nicely get rid of `Theme`. But we can reduce its unfriendliness by 1) creating an alias called `Style` which seems to be a more well-known name for this sort of thing 2) choosing a different color for each layer automatically. That said, here's the pattern I see coming from option 3:

1.
```julia
plot(iris, x = :SepalWidth, y = :PetalLength, layer(Geom.point),  layer(Geom.smooth))
```
[...]

PR #1459, opened 2020-06-21
Geom.line update

### LINKED ISSUE #1391: Weird stacking of layers
The code is complicated so I cannot give you a simple breaking example. However, in essence, I plot a series of gray curves and a single red curve. However, when the curves are plotted the red curve is sometimes behind and sometimes in front of the gray layers. How is this possible?!

image

the code is something like this:

```
colormap =
function(nc)
[Gadfly.parse_colorant(["red"]); repeat(Gadfly.parse_colorant(["gray"]), inner=nc-1)]
end
pS = Gadfly.plot(X, x=repeat(xaxis, inner=nS), y=Gadfly.Col.value(1:nS...), color=colindex, group=colindex, Gadfly.Scale.color_discrete(colormap))
```
What I can potentially do to avoid this from hqppening?!

Mattriks:

### This PR:
-  updates `Geom.line` (following GiovineItalia/Compose.jl#397, fixes #1391)
- `Geom.line` supports one-length arrays e.g. `linestyle=[:dash]` or `color=[colorant"red"]`  (following #1438). This issue has been mentioned e.g. https://github.com/GiovineItalia/Gadfly.jl/issues/818#issuecomment-480466137

### To be done:
- [x] bump compat for Compose.jl (needed for this PR)
- [x] in the Gadfly docs, replace the syntax `layer(..., Geom.line, Theme(...))` where possible

### Example:
```julia
y1 = [0.1, 0.26, missing, 0.5, 0.4, NaN, 0.48, 0.58, 0.83]

plot(x=1:9, y=y1, Geom.line, Geom.point, color=["Item 1"], linestyle=[:dash], size=[3pt],
    layer(x=:1:10, y=rand(10), Geom.line, Geom.point, 
        color=["Item 2"], size=[5pt], shape=[Shape.square]),
    layer(x=:1:10, y=rand(10), color=[colorant"hotpink"], 
        linestyle=[[8pt, 3pt, 2pt, 3pt]], Geom.line))
```
geom_line


Mattriks:
I did the regression testing, and the main differences between master (with Compose 0.8.2) and dev (with Compose 0.9) are not due to this PR (rather GiovineItalia/Compose.jl#392). The main differences were in the tick labels for transformed axes with supercript exponents: now the tick labels are better spaced vertically (particularly for the x-axis).

Some plots have a trivial difference in line order due to this PR (i.e. whether a line is drawn above or below another), but nothing else about the plot is changed.

PR #1467, opened 2020-07-28
Update instructions for using PackageCompiler.jl

f-ttok:
I found that [PackageCompiler.jl](https://github.com/JuliaLang/PackageCompiler.jl) had been updated and procedures to compile packages had been changed.
This PR gives some modifications to instructions to compile Gadfly.jl in the document.

In my environment, PackageCompiler.jl added by `]add PackageCompiler` (not `]dev PackageCompiler` )  works well, so I think the first caveat may not be necessary.

> PackageCompiler is a work in progress, and the most recent tagged release
> sometimes (currently as of this writing) does not work.  Hence the `dev`
> instead of `add` command to install.


bjarthur:
thanks!  looks great.

Issue #1466, opened 2020-07-18
Incompatibility with JuliaImage ImageCore.jl


NicholasWMRitchie:
Gadfly fails when manual_color_key(...) is used when ImageCore is loaded

```julia
using Gadfly
using Colors

curves = 1:3

const palette = distinguishable_colors(
    length(curves)+3,
    [ RGB(253 / 255, 253 / 255, 241 / 255), RGB(0, 0, 0), colorant"DodgerBlue4"],
    transform = deuteranopic,
)[3:end]


ls = [layer(x=1:10, y=map(x->x^i, 1:10), Geom.line, Theme(default_color = palette[i] )) for i in curves ]
# Succeeds
plot(ls..., Guide.manual_color_key("Curves", [ "Curve $i" for i in curves ], palette[curves]))

using ImageCore
# Fails on the second plot...
plot(ls..., Guide.manual_color_key("Curves", [ "Curve $i" for i in curves ], palette[curves]))
```

The problem is the line in [ImageCore/convert_reinterpret.jl](https://github.com/JuliaImages/ImageCore.jl/blob/master/src/convert_reinterpret.jl) line 80-83

```julia
function Base.convert(::Type{Array{Cdest,n}},
                      img::AbstractArray{Csrc,n}) where {Cdest<:Colorant,n,Csrc<:Colorant}
    copyto!(Array{ccolor(Cdest, Csrc)}(undef, size(img)), img)
end
```

This leads to this error:
```
ERROR: LoadError: TypeError: in new, expected Array{Colorant,1}, got a value of type Array{RGB{Float64},1}
Stacktrace:
 [1] ManualDiscreteKey at C:\Users\nritchie\.julia\dev\Gadfly\src\guide\keys.jl:267 [inlined]
 [2] Gadfly.Guide.ManualDiscreteKey(; title::String, labels::Array{String,1}, pos::Array{Any,1}, color::Array{RGB{Float64},1}, shape::Array{Function,1}, size::Array{Measures.Measure,1}) at C:\Users\nritchie\.julia\dev\Gadfly\src\guide\keys.jl:295
 [3] #manual_color_key#131 at C:\Users\nritchie\.julia\dev\Gadfly\src\guide\keys.jl:321 [inlined]
 [4] manual_color_key(::String, ::Array{String,1}, ::Array{RGB{Float64},1}) at C:\Users\nritchie\.julia\dev\Gadfly\src\guide\keys.jl:321
[...]

Issue #1454, opened 2020-06-13
manual_color_key fails when handed color = Colorant[RGB(...), RGB(...)]


NicholasWMRitchie:
This fails on version 1.3.0 but not 1.2.1

    Guide.manual_color_key("Title",["Item 1","Item 2"], Colorant[RGB(1.0,0.0,0.0),RGB(0.0,1.0,0.0)])

with
    ERROR: TypeError: in new, expected Array{Colorant,1}, got Array{RGB{Float64},1}

    
I think the problem is `parse_color(...)` in `Compose`

    parse_colorant(c::Union{Tuple,Vector}) = [parse_colorant(x) for x in c]

which should be

    parse_colorant(c::Union{Tuple,Vector}) = Colorant[parse_colorant(x) for x in c]

In the meanwhile, adding 
    
    Compose.parse_colorant(c::Array{<:Colorant,1}) = convert(Array{Colorant,1}, c)

resolves the issue without modifying `Compose`

NicholasWMRitchie:
Probably a better way to handle this would be to change `manual_color_key` to take a

    color::Array{<:Colorant,1}

rather than a

    color::Colorant[]

Mattriks:
I'm not seeing this error. What version of Colors and Compose do you have (I have Colors 0.12.2 and Compose#master)  ?:
```julia
p = plot(
    Guide.manual_color_key("Title",["Item 1","Item 2"], [RGB(1.0,0.0,0.0),RGB(0.0,1.0,0.0)],
        pos=[0.5, 1.5])
)
draw(PNG(), p)
```
![issue1454](https://user-images.githubusercontent.com/18226881/84568268-83d09600-adc1-11ea-8066-f51a167d60a9.png)

NicholasWMRitchie:
v0.8.2

> 
> 
> I'm not seeing this error. What version of Colors and Compose do you have (I have Colors 0.12.2 and Compose#master) ?:
> 
> ```julia
> p = plot(
>     Guide.manual_color_key("Title",["Item 1","Item 2"], [RGB(1.0,0.0,0.0),RGB(0.0,1.0,0.0)],
>         pos=[0.5, 1.5])
> )
> draw(PNG(), p)
> ```
> 
> ![issue1454](https://user-images.githubusercontent.com/18226881/84568268-83d09600-adc1-11ea-8066-f51a167d60a9.png)
> 
[...]

Issue #1342, opened 2019-11-08
StackOverflowError: Gadfly.draw() fails for large images


montyvesselinov:
```
(v1.2) pkg> status Gadfly
    Status `~/.julia/environments/v1.2/Project.toml`
  [5ae59095] Colors v0.9.6
  [34da2185] Compat v2.2.0
  [a81c6b42] Compose v0.7.4 #master (https://github.com/GiovineItalia/Compose.jl.git)
  [864edb3b] DataStructures v0.17.5
  [31c24e10] Distributions v0.21.6
  [c91e804a] Gadfly v1.0.1
  [c8e1da08] IterTools v1.2.0
  [682c06a0] JSON v0.21.0
  [e5e0dc1b] Juno v0.7.2
  [442fdcdd] Measures v0.3.0
  [ae029012] Requires v0.5.2
  [2913bbd2] StatsBase v0.32.0
  [ade2ca70] Dates
  [8dfed614] Test
```

```
X=rand(50,10000);
pS = Gadfly.plot([Gadfly.layer(x=1:50, y=X[:,i], color=["c$(i)" for j in 1:50], Gadfly.Geom.line(), Gadfly.Theme(line_width=2Gadfly.pt)) for i=1:10000]..., Gadfly.Theme(key_position=:none))
```

this works.

```
display(pS)
```

also works. However,

```
Gadfly.draw(Gadfly.PDF("a.pdf"), pS)
Gadfly.draw(Gadfly.PNG("a.png"), pS)
```

fail:

```
ERROR: StackOverflowError:
Stacktrace:
 [1] compose!(::Compose.Context, ::Compose.Context, ::Compose.Context, ::Compose.Context, ::Vararg{Compose.Context,N} where N) at /Users/monty/.julia/packages/Compose/6FIbD/src/container.jl:310 (repeats 9620 times)
[...]

PR #1470, opened 2020-08-10
force type `Vector{Colorant}` for `ManualDiscreteKey`

dehann:
Fixes an unexpected dispatch problem with a minimal code change. Seemed like the best solution after various attempts to fix downstream with user code, but didn't work.  This issue also seems to be a regression since the user code has been working for quite some time.

```julia
ERROR: LoadError: LoadError: TypeError: in new, expected Array{Colorant,1}, got a value of type Array{RGB{Normed{UInt8,8}},1}
Stacktrace:
 [1] Gadfly.Guide.ManualDiscreteKey(::String, ::Array{String,1}, ::Array{Any,1}, ::Array{RGB{Normed{UInt8,8}},1}, ::Array{Function,1}, ::Array{Measures.Measure,1}, ::Bool) at /home/dehann/.julia/dev/Gadfly/src/guide/keys.jl:267
 [2] Gadfly.Guide.ManualDiscreteKey(; title::String, labels::Array{String,1}, pos::Array{Any,1}, color::Array{String,1}, shape::Array{Function,1}, size::Array{Measures.Measure,1}) at /home/dehann/.julia/dev/Gadfly/src/guide/keys.jl:295
 [3] #manual_color_key#131 at /home/dehann/.julia/dev/Gadfly/src/guide/keys.jl:321 [inlined]
 [4] manual_color_key at /home/dehann/.julia/dev/Gadfly/src/guide/keys.jl:321 [inlined]
 [5] plotKDE(::Array{BallTreeDensity,1}; c::Array{String,1}, N::Int64, rmax::Float64, rmin::Float64, axis::Nothing, dims::Nothing, xlbl::String, title::String, legend::Array{String,1}, dimLbls::Nothing, levels::Int64, fill::Bool, points::Bool, layers::Bool, overlay::Nothing) at /home/dehann/.julia/packages/KernelDensityEstimatePlotting/49JcX/src/KernelDensityEstimatePlotting.jl:356
```

Not sure if this is related, but:
```julia
julia> using Colors
julia> RGB isa Colorant
false
julia> RGB <: Colorant
true

```

# Contributor checklist:

- [ ] N/A I've updated the documentation to reflect these changes
- [ ] N/A I've added and/or updated the unit tests

# Proposed changes

- force type `Vector{Colorant}`for `ManualDiscreteKey` in guide/keys.jl

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@ Each release typically has a number of minor bug fixes beyond what is listed her
  * Better define what scales are included in the (internal) `scales` dict (#1468)
  * Support one-length aesthetics for `Geom.segment` (#1465)
  * Support one-length aesthetics e.g. `color=[colorant"red"]` for `Geom.line` (#1459)
+ * Fix dispatch problem on `ManualDiscreteKey` as `Vector{Colorant}` (#1470)
 
 
 
--- a/src/guide/keys.jl
+++ b/src/guide/keys.jl
@@ -292,7 +292,7 @@ function ManualDiscreteKey(;title="", labels=String[], pos=[], color=Colorant[],
         swatches = collect(Tuple, zip(cataes[notempty]...))
         !allunique(swatches) && error("Swatches should not be repeated in a manual key")
     end
-    return ManualDiscreteKey(title, labels, pos, clrs, shps, szs,  true)
+    return ManualDiscreteKey(title, labels, pos, Vector{Colorant}(clrs), shps, szs,  true)
 end
 
 


Mattriks:
Is this due to a type piracy issue by another package like in #1466? i.e. Can you demonstrate that the above error is caused by using Gadfly alone, without some other package being the root cause?

dehann:
Hi ah, this is likely the same problem then:
- #1466,
- #1454 
i am also loading Images.jl.  So maybe best to wait for JuliaImages/ImageCore.jl#131 to be merged.  

This PR should be pretty harmless, since it only changes colors object type a little earlier in the stack, rather than an automated `convert` when `new` is called.  I think what's happening is this PR is calling the convert inside the `Gadfly` context, which avoids the user code from calling the convert function in `Main` context, which opens the door for outside converters that don't dispatch properly to cause the error.

Mattriks:
Can you give an example of your plot which uses `Guide.manual_color_key`.  There may be a more automated way to do the plot now with recent developments on Gadfly#master (e.g. #1459)

dehann:
on Gadfly#master, but doesn't error without Images.
```julia
using Gadfly, Images
Guide.manual_color_key("Legend", ["red"; "blue"], parse.(Colorant, ["red";"blue"]))

ERROR: TypeError: in new, expected Array{Colorant,1}, got a value of type Array{RGB{Normed{UInt8,8}},1}
Stacktrace:
 [1] ManualDiscreteKey at /home/dehann/.julia/packages/Gadfly/asb1x/src/guide/keys.jl:267 [inlined]
 [2] Gadfly.Guide.ManualDiscreteKey(; title::String, labels::Array{String,1}, pos::Array{Any,1}, color::Array{RGB{Normed{UInt8,8}},1}, shape::Array{Function,1}, size::Array{Measures.Measure,1}) at /home/dehann/.julia/packages/Gadfly/asb1x/src/guide/keys.jl:295
 [3] #manual_color_key#131 at /home/dehann/.julia/packages/Gadfly/asb1x/src/guide/keys.jl:321 [inlined]
 [4] manual_color_key(::String, ::Array{String,1}, ::Array{RGB{Normed{UInt8,8}},1}) at /home/dehann/.julia/packages/Gadfly/asb1x/src/guide/keys.jl:321
 [5] top-level scope at REPL[16]:1
```

dehann:
An alternative suggestion here might be to force the express field type `::Vector{Colorant}` with a bespoke inner constructor that allows the broader `::Vetor{<:Colorant}`, until the type piracy in ImageCore.jl is fixed:
```julia
ManualDiscreteKey(title, labels, pos, clrs::Vector{<:Colorant}, shps, szs,  vis) = new(title, labels, pos, Vector{Colorant}(clrs), shps, szs,  vis)
```

This might be more fitting than only 'fixing' the type inside the convenience constructor `manual_color_key(..)`.  Just a thought.

Mattriks:
Thanks for the extra info about your style of plot.  It is possible to automate the colorkey, and we are working to further increase support for this new style syntax. Here are some examples:
- #1459
- #1475
- a Geom.density example in the plot gallery ([3rd plot](http://gadflyjl.org/dev/gallery/geometries/#[Geom.density](@ref)-1))
- wide-format style #1342

Feedback welcome!

dehann:
Thanks, will work through the examples you listed, and more automation on colorkey is great news!  Thanks again.

Regarding the type piracy problem from ImageCore.jl, perhaps we should leave this PR open so others can find it more easily, but then close this PR when JuliaImages/ImageCore.jl#131 is merged and tagged?

dehann:
Hi, I just tried this again and looks to be resolved (following recent ImageCore.jl tag).  I tried with latest `Gadfly#master` and with updated packages.  I will close 

Consequently @Mattriks , perhaps I could ask for a new tag on Gadfly.jl if possible please?  We are having Pkg compat issues downstream which likely follow from this type piracy problem -- we think that a release new tag on Gadfly will help resolve the problem.

cc @Affie

Issue #1482, opened 2020-09-26
Perhaps tag new patch version?


dehann:
Hi, would it perhaps be possible to tag patch version v1.3.1 please?

EDIT: following fix to #1466

dehann:
Hi, a friendly bump on this issue?

ChrisRackauckas:
Friendly bump? This was identified as the source of an issue on the Slack and Discourse. The DataStructures 0.18 compatibility commit is particularly necessary to get tagged.

dehann:
Hi @bjarthur , thank you very much!

PR #1484, opened 2020-10-11
tag v1.3.1

bjarthur:
closes https://github.com/GiovineItalia/Gadfly.jl/issues/1482


bjarthur:
note that CI on 1.0.5 fails because Cairo now requires 1.3

PR #1499, opened 2020-11-18
Fix tests with DataFrames 0.22 on Julia 1.5

nalimilan:


rikhuijzer:
Note that if you run the CI after upgrading CategoricalArrays so that DataFrames goes to 0.22, one test will fail (https://github.com/GiovineItalia/Gadfly.jl/runs/1417459612).

nalimilan:
Thanks. I've added a fix for that test. But CI still uses DataFrames 0.21.8 (not sure why since I've just updated RDatasets to allow 0.22, and CSV doesn't depend on DataFrames).

rikhuijzer:
That other PR is mostly working on fixing the failing Julia 1.0 (LTS). What happens if you bump RDatasets? (Note that I can't do it for you because I'm not an official maintainer here at Gadfly. I just saw something at my PR which might help your PR out.)

nalimilan:
Tests now pass with DataFrames 0.22 on Julia 1.5.

Issue #1272, opened 2019-04-11
Zooming in IJulia notebooks


isentropic:
I'm failing to use the mouse zooming in IJulia
Specifically, panning works with no problem, but when I try to zoom it (shift+leftclick), it zooms to a seaming random location (off the limits from the current plot)

However, it works if the plot was made from REPL (standalone html with the plot works ok)

Here are some logs:
```

(v1.1) pkg> status
    Status `~/.julia/environments/v1.1/Project.toml`
  [7e558dbc] ArbNumerics v0.4.5
  [6e4b80f9] BenchmarkTools v0.4.2
  [336ed68f] CSV v0.4.3
  [40e3b903] Clang v0.9.0
  [a81c6b42] Compose v0.7.3
  [a0b5b9ef] Cxx v0.2.0+ [`~/.julia/dev/Cxx`]
  [1f15a43c] CxxWrap v0.8.1
  [717857b8] DSP v0.5.2
  [a93c6f00] DataFrames v0.17.1
  [0c46a032] DifferentialEquations v6.3.0
  [31c24e10] Distributions v0.17.0
  [d91b1130] EllipseFit v0.0.0 #master (https://github.com/isentropic/EllipseFit.jl)
  [7a1cc6ca] FFTW v0.2.4
  [c91e804a] Gadfly v1.0.1
  [4b11ee91] Gaston v0.9.2
  [7073ff75] IJulia v1.18.0
  [a98d9a8b] Interpolations v0.11.2
  [2b0e0bc5] LanguageServer v0.5.1
  [9b87118b] PackageCompiler v0.6.3
  [91a5bcdd] Plots v0.24.0
  [c46f51b8] ProfileView v0.4.0
  [438e738f] PyCall v1.91.1
  [d330b81b] PyPlot v2.8.1
  [90137ffa] StaticArrays v0.10.3

(v1.1) pkg>
```
and 
```

/mnt/BE7C43F57C43A74D/PROJECTS » julia                               zhanibek@zhanibek-MS-7A34
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.1.0 (2019-01-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |


```

[...]

PR #1505, opened 2020-12-09
Zoom fixes

### LINKED ISSUE #1504: Shift + scrolling to zoom in Firefox results in NaN values
On the latest version of Firefox, shift+scrolling causes the co-ordinates to be set to NaN, and plot interactivity stops working. 

This seems related to the discourse link posted in #1272 

I have an incoming pull request that tries to fix this.

blepabyte:
Fixes #1504 

# Contributor checklist:

# Proposed changes

- The `mousewheel` and `DOMMouseScroll` events are deprecated (`wheelDelta` was undefined causing the NaN's) so I swapped to the `wheel` event, which has [better](https://caniuse.com/?search=wheel) browser support. 
- Made some changes to address the docs issue in #1272 - I built the docs and zoom is now working on all the browsers I tested (additionally, the co-ordinates are now correct)
    - Plotting from REPL worked since the plot was always at the exact top-left corner of the page

This does not completely fix the original IJulia notebook problem in the linked issue (when shift left-click + dragging to zoom) as jupyter seems to do something weird with mouse events.

### DIFF
--- a/src/gadfly.js
+++ b/src/gadfly.js
@@ -110,9 +110,7 @@ Snap.plugin(function (Snap, Element, Paper, global) {
             };
         }
 
-        this.node.addEventListener(
-            /Firefox/i.test(navigator.userAgent) ? "DOMMouseScroll" : "mousewheel",
-            fn2);
+        this.node.addEventListener("wheel", fn2);
 
         return this;
     };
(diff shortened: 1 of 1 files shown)


bjarthur:
thanks!

PR #1519, opened 2021-02-18
Added ability for intercept in geom.hair to take vector to function as lolipop

hewsond:
Chainged Geom.hair
Documented hair as lolipop in geometries documentation
Added Geom.hair update as lolipop to news
Added tests for Geom.hair lolipop

Fixes #XXX

# Contributor checklist:

# Proposed changes
Use length check on `intercept` in `Geom.hair` to determine if it is same length as number of points want to draw hairs for.

If it is same length, don't fill end point vector, just use the supplied vector. 

hair

# Doc building issues
I've added it to the test files and above you can see the output, however when i try to build the documentation I get an error:
 `MethodError: no method matching isless(::Array{Float64,1}, ::Int64)`

Though it isn't the only `@example` block that gives some errors.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@ Each release typically has a number of minor bug fixes beyond what is listed her
  * Add "vertical" orientation for `Geom.ribbon` (#1513)
  * Support one-length aesthetics for `Geom.polygon` and `Geom.ribbon` (#1511)
  * Enable `color` grouping for `Geom.density2d` (#1508)
+ * Add ability for `intercept` to be a vector in `Geom.hair` to function as a lolipop chart (#1518)
 
 # Version 1.3.1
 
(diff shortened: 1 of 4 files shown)


Mattriks:
Another possibility here is changing the syntax to `plot(x=x, y=x.^2, intercept=(x.^2)./2, Geom.hair, Geom.point)`, and allow `intercept` to take vectors of length one (or more).  I've got a PR coming soon which refactors some internal code and is relevant to making that switch.

hewsond:
I'm not an expert on the whole grammar of graphics thing but IMO the intercept argument as you suggest does sound much nicer and cleaner.

I'm happy to have at re-basing to work in that way once you've done the refactor if required.

bjarthur:
@Mattriks did you finished the aforementioned PR?

Mattriks:
No unfortunately.  I think I was referring to #1520 (which I opened on the same day as the posts above), or a subsequent PR.  I do hope to return to #1520 sometime!

PR #1520, opened 2021-02-19
Make plot layers & inner subplot layers consistent

### LINKED ISSUE #1516: Stats in Geom.subplot_grid
In the main Gadfly plot, Stats and Geoms that are not in layers, are both pushed into a layer (which makes sense). This means the `statistics` field of the main plot (`p1.statistics` below) is actually obsolete. 

```julia
p1 = plot(Stat.smooth, Geom.ribbon)
p1.statistics #  StatisticElement[]
 p1.layers[1].statistics # StatisticElement[Stat.smooth()]
 p1.layers[1].geom # Geom.ribbon() 
```
</br>

`Geom.subplot_grid` has similar fields to the main plot. In `subplot_grid` Geoms that are not in layers get pushed to a layer,
but Stats get pushed to the `statistics` field of `subplot_grid` (`p2.statistics` below). This is problematic, 
because the Stat and Geom don't have a clear association (this is worse when attempting to use multiple Stats/Geoms).

```julia
p2 = Geom.subplot_grid(Stat.smooth, Geom.ribbon)
p2.statistics  # StatisticElement[Stat.smooth()]
 p2.layers[1].statistics # StatisticElement[]
 p2.layers[1].geom # Geom.ribbon()
```

The decoupling of the Stat/Geom makes it makes it unclear what the purpose of `p2.statistics` is, leading to bad coding.

Mattriks:

### This PR:

- makes the `layer.statistics` field of `plot(layer(...))` and the inner layers of `Geom.subplot_grid()` consistent (fixes #1516)
- updates `Stat.violin` as needed
- an open question from this PR is what should be in the `layer.statistics` field of the _outer_ layer of `layer(Geom.subplot_grid(layer(...)))`. Note that this should not necessarily be the same as 
the `statistics` field of the inner layers, which are clearly defined by this PR.  
E.g. the outer layer of `Geom.subplot_grid` could contain statistics which modify aesthetics such as `:xgroup` and `:ygroup`, but
no such statistics have been developed yet for Gadfly.

### Example 
(now `p2` shows same behaviour as `p1` from OP in #1516):
```julia
p2 = Geom.subplot_grid(Stat.smooth, Geom.ribbon)
p2.statistics   # StatisticELement[]
p2.layers[1].statistics # StatisticElement[Stat.smooth()]
```

- Note in this PR, the fields `p1.statistics` and `p2.statistics` have not been deprecated, because they may be useful for developers.

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -182,7 +182,12 @@ function layer(data_source,
     for element in elements
         add_plot_element!(lyrs, element)
     end
-    lyrs
+    for l in lyrs
+        if isempty(l.statistics) && !isa(l.geom, Geom.Nil) && !isa(l.geom, Geom.SubplotGeometry)
+            l.statistics = [default_statistic(l.geom)]
+        end
+    end
+    return lyrs
 end
 
 """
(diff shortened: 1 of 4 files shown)


Mattriks:
Regression testing showed 1 difference: testscripts/auto_enumerate.jl shows `Geom.line` layer but no `Geom.bar`. Will investigate.

PR #1530, opened 2021-04-01
make Gadfly "relocatable" by not loading files at runtime

KristofferC:
Gadfly currently loads two source files at runtime using Requires to add extra functionality when some packages are loaded. One issue with this is that if you compile Gadfly into a sysimage and move the sysimage somewhere else, these files will not exist and Gadfly will not work. An example of when this happens is when using PackageCompiler (https://github.com/JuliaLang/PackageCompiler.jl/issues/498).

Requires.jl has since https://github.com/JuliaPackaging/Requires.jl/pull/85 the capability of storing the source code that would be included inside the package. When the `@requires` block runs, there is then no need to actually read the file for the code, instead, the code saved inside the package is used. However, this functionality only works if the `@requires` block is a simple `include` statement. Therefore, this PR does the small changes to allow this to work and allowed Gadfly to be used with sysimage and not have those sysimages depend on paths on the local machine.

### DIFF
--- a/src/Gadfly.jl
+++ b/src/Gadfly.jl
@@ -41,11 +41,6 @@ export Plot, Layer, Theme, Col, Row, Scale, Coord, Geom, Guide, Stat, Shape, ren
 export SVGJS, SVG, PGF, PNG, PS, PDF, draw, inch, mm, cm, px, pt, color, @colorant_str, vstack, hstack, title, gridstack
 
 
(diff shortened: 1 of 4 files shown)


bjarthur:
LGTM.  @Mattriks ?  

and @KristofferC -- thanks so much for all the work you've put into PackageCompiler.  it's a joy to use now!

Issue #1532, opened 2021-04-24
Text is overlapping/improperly cropped


JarredAllen:
When I try to make any plots in a jupyter notebook using Gadfly, or when I look at the documentation, the text on the plots is messed up. Text overlaps other text, and some text gets cropped out of the plot when it shouldn't be.

For example, here is what I see when I go to [http://gadflyjl.org/stable/man/compositing/#Facets](http://gadflyjl.org/stable/man/compositing/#Facets):
![image](https://user-images.githubusercontent.com/19656507/115973086-5f783680-a507-11eb-88c6-e2f718d26c67.png)
I'd guess that the x-axis label is supposed to be "SepalLength by Species", but the end is cropped out of the image, and the beginning overlaps with the y-axis label. Other text in the plot is similarly messed up.

If this is relevant, I'm using Firefox 87.0 on Ubuntu 20.04.

Mattriks:
Chrome renders the text correctly, so this looks like a FF issue e.g. google "firefox svg text too big".
Compose.jl (the Graphics package underlying Gadfly) for svg uses unitless values ("user units" set by the viewBox)  similar to the first example shown [here](https://mpetroff.net/2013/08/analysis-of-svg-units/). So font-size is unitless in svgs drawn by Compose.

You can test what FF is doing by creating a simple plot:
```julia
p = plot(x=1:5, y=rand(5), Geom.point)
draw(SVG("test1.svg"), p)
```
Open "test1.svg" in a text editor, and add e.g. `font-size=3.88px` on line 18, and open the svg file in FF.  Are unitless font-sizes a problem for FF?

JarredAllen:
It was the minimum font size setting. It seems that firefox defaulted to having it be 12pt font (at least for me, since I'm pretty sure I hadn't touched the advanced font settings), which was messing up the rendering. I've disabled that setting and now it seems to work normally.

PR #1534, opened 2021-04-26
Add note to tutorial.md explaining pain point

JarredAllen:
Add note explaining that a minimum font size in your browser can cause improper text rendering (as happened to me in issue #1532 the other day).

Hopefully this will help others avoid/resolve this pain point that I hit.

# Contributor checklist:

Most of these items don't apply since I'm just proposing a minor addition to the documentation

# Proposed changes

- Minor addition to documentation in tutorial (explained above)

### DIFF
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -73,6 +73,10 @@ Now we have the following charming little SVG image.
 p # hide
 ```
 
+If some of the text in this image is overlapping other text, your browser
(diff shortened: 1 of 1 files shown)


JarredAllen:
Are there any updates on the consideration of this? It would have improved my experience greatly when initially trying to use this library if this change was present.

bjarthur:
thanks!

PR #1536, opened 2021-04-29
Update CompatHelper.yml

rikhuijzer:
This is a copy of the newest CompatHelper configuration from their [documentation](https://juliaregistries.github.io/CompatHelper.jl/dev/). Note that, here, the `COMPATHELPER_PRIV` key is also set meaning that CI will run for pull requests created by CompatHelper.

Also, this adds the `workflow_dispatch` which is convenient for package maintainers. With this, you can run CompatHelper from the [Actions tab](https://github.com/GiovineItalia/Gadfly.jl/actions).


rikhuijzer:
Thanks

PR #1562, opened 2021-12-28
fix interactive hiding of ribbons

bjarthur:
clicking on the color swatch in the key now turns off ribbons too, just like it always has for lines.


tlnagy:
Thanks for fixing this! Always bothered me when it didn't switch off.

Issue #1271, opened 2019-04-05
Display data values under cursor when using SVGJS


robsmith11:
A very useful feature with ggplot2/plotly is that when the mouse cursor is placed over a data point, it's exact values with be interactively displayed (coordinates, aesthetics, geom_text values, etc.).  It would be nice to have this feature when using Gadfly with SVGJS.

Mattriks:
Did you read the interactivity blurb in the tutorial:
http://gadflyjl.org/dev/tutorial/#Interactivity-1

All the plots in the docs have this functionality: they are rendered in SVGJS.

robsmith11:
Are you referring to the cursor coordinates that can be shown by pressing `c`?

That's a bit different than displaying the exact values of nearest data point for a couple reasons:
1.  Often one wants to know the values of a specific data point with full precision, not the rounded value of a nearby pixel.
2.  It's useful to interactively display additional arbitrary text for the chosen data point that would otherwise clutter the plot.

bjarthur:
agreed.  i would happily review a PR.

PR #1563, opened 2021-12-29
add interactive datatips for Geom.point

bjarthur:
fixes https://github.com/GiovineItalia/Gadfly.jl/issues/1271


bjarthur:
@robsmith11 care to try this out?

PR #1591, opened 2022-06-11
Remove export of non-existent prepare_display

### LINKED ISSUE #1578: Why prepare_display is still exported?
`prepare_display` is exported from `Gadfly` but there's no method definitions.
https://github.com/GiovineItalia/Gadfly.jl/blob/423f170cc5fa4966f1dfc31640464e02c2e4e588/src/Gadfly.jl#L25
Is it expected that this may be useful for clients somehow (how?) or is it simply an oversight?

ulysses4ever:
Fixes #1578


bjarthur:
thanks!

PR #1601, opened 2022-10-28
handle dates properly in interactive SVG coordinates

bjarthur:
@dcjones per your [comment](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/ticks.jl#L311) about improving multi-level ticks for DateTime, i have taken a stab at refactoring `optimize_ticks(x_min::DateTime, ...` to use something akin to the Wilkinson's ad-hoc scoring method currently used for all other types.  could you please review?

one specific question i have is the `extend_ticks` keyword argument.  when `true` it widens the range on both sides thereby tripling the number of ticks.  similarly, `multilevel_ticks` is called with a [tripled range](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/statistics.jl#L971-L972).  why?  `extend_ticks` is set to `true` in [src/ticks.jl](https://github.com/GiovineItalia/Gadfly.jl/blob/master/src/statistics.jl#L964) but left to the default `false` in src/stats.  if i set `extend_ticks` to `false` in ticks.jl, and remove the widening of the range when calling `multilevel_ticks`, none of the regressions tests are different in any meaningful way.

examples of zooming in on axes with Dates now in interactive SVG follow.

quarterly:

Screen Shot 2022-11-20 at 11 38 05 AM

monthly:

Screen Shot 2022-11-20 at 11 38 21 AM

weekly:

Screen Shot 2022-11-20 at 11 39 34 AM

the latter also shows a datatip with a date.


bjarthur:
i think this PR is ready to merge.  @dcjones can you please review?  git-blame says you were the last to touch the `extend_ticks` code which i ask about in the edited OP above.  thanks.

dcjones:
It's been a while since I've looked at this code, but this looks like a nice improvement. I did run into one issue. This isn't really handled by the existing code either, but because time period types are integers, there can be errors when trying to multiply them by a non-integer step size, which can happen when optimizing ticks.

Here's a test case that triggers that.

```julia
using Gadfly, Dates

xs = Millisecond(1):Millisecond(2):Millisecond(200)
ys = randn(length(xs))

plot(x=xs, y=ys, Geom.line) |> SVG("test.svg", 4inch, 4inch)
```

Not sure what the simplest solution is, but somehow step sizes need to be restricted to integers when using `TimePeriod` types.

bjarthur:
thanks.  it's seems like what we need then is an `optimize_ticks(x_min::Period, ...` method to address your issue.  i'm inclined to just merge this PR as is and leave that for another one.

i'm still curious thought about `extend_ticks`.  i don't see why one would ever want to set that to `true`.

dcjones:
I think the original intention for the extend ticks thing was for use with interactive SVG, so that you could zoom out and have ticks marks that extend beyond the normal range, which are visible only when zoomed out. I guess it's debatable how useful that is, and it doesn't appear to be used anyway.

bjarthur:
thanks for the info.  i guess i'll leave `extend_ticks` alone for now too.

the other thing related to this PR that i think i'll also save for later, is to somehow make `scales` in `multilevel_ticks` user configurable.  perhaps through `Theme` or an argument into `Stat.ticks`.

Issue #1589, opened 2022-06-03
Alpha for Geom.line and Geom.path


arnold-c:
Apologies if this is covered elsewhere, but I couldn't find it in the issues or the documentation. Is it possible to enable `Theme(alphas = [])` for `Geom.line` and `Geom.path`? It would be very helpful for situations where trajectories are simulated and are overlayed e.g. stochastic disease modeling. Small reprex below (`alphas` works as expected when the geom is changed to `Geom.point`, for example).

Thanks very much

```
test = DataFrame(
  x = 1:100,
  y = rand(100)
)

Gadfly.plot(
  test,
  x = :x,
  y = :y,
  Geom.line,
  Theme(alphas = [0.1])
)
```

Rapsodia86:
This is not a direct solution you are asking for, but maybe will help. 
See: https://discourse.julialang.org/t/gadfly-how-to-change-width-and-opacity-of-lines-in-geom-density/37823/2
Using Colors library, you can set a style:

```
using Gadfly
using Colors

test = DataFrame(
    x=1:100,
    y=rand(100)
)

Gadfly.plot(
    test,
    x=:x,
    y=:y,
    Geom.line, style(
        default_color=RGBA(0.1, 0.1, 0.9, 0.3))
)


```

arnold-c:
Thanks @Rapsodia8. Sorry for the follow up - I should have made a better reprex. I have more than one group that produce the colors. From the documentation it seems like I could use `plot(Scale.color_discrete_manual())` to set the colors, but I've only got it working when naming the colors. Is there a way to use `RGBA()` (and therefore specify the alphas)?

```
using Gadfly, Colors

test = DataFrame(
    x=1:100,
    y=rand(100),
    z = repeat(["a", "b"], outer = 50)
)

personal_theme = Theme(
  panel_fill="white",
  background_color = "white"
  )

Gadfly.plot(
    test,
    x=:x,
    y=:y,
    color = :z,
    Geom.line,
    personal_theme
    Scale.color_discrete_manual("red", "blue")
)
```

arnold-c:
Thanks for your help. I was receiving the same error. I have updated the code to plot each series in a `for` loop. Code below in case it's of any help to others/please feel free to point out bad practices given my inexperience with Julia.

```
using Gadfly, Colors, DataFrames, DataFramesMeta

test = DataFrame(
    x = 1:100,
    y = rand(100),
[...]

PR #1603, opened 2022-12-11
add alpha aesthetic to Geom.line

bjarthur:
fixes https://github.com/GiovineItalia/Gadfly.jl/issues/1589

```
julia> using Gadfly, DataFrames

julia> test = DataFrame(
           x=1:100,
           y=rand(100),
           z = repeat(["a", "b"], outer = 50));

julia> plot(test,
            x=:x,
            y=:y,
            color = :z,
            Geom.line,
            alpha=[0.1])

julia> plot(test,
            x=:x,
            y=:y,
            alpha = :z,
            Geom.line,
            Theme(alphas=[1.0,0.5]))
```
Screen Shot 2022-12-11 at 10 59 25 AM

Screen Shot 2022-12-11 at 10 59 32 AM

### DIFF
--- a/src/geom/line.jl
+++ b/src/geom/line.jl
@@ -20,8 +20,8 @@ end
     Geom.line[(; preserve_order=false, order=2)]
 
 Draw a line connecting the `x` and `y` coordinates.  Optionally plot multiple
-lines according to the `group`, `color` or `linestyle` aesthetics.  `order` controls whether
-the lines(s) are underneath or on top of other forms.
+lines according to the `group`, `color`, `alpha`, or `linestyle` aesthetics.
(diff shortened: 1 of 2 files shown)


Mattriks:
I'm surprised that this PR works (https://github.com/GiovineItalia/Compose.jl/issues/386#issuecomment-586451902), but nice that it does.  I will try to think of some examples that it might fail for (due to Compose), but I think PR is ok as is (we can fix any issues in Compose later).

bjarthur:
i just added a test.  SVG, PDF, and PNG output all work for me.  @Mattriks let me know if you find a case if fails.

Issue #1607, opened 2023-02-20
tests and building docs currently failing


bjarthur:
specifically subplot_grid_guides.jl.  stacktrace indicates a problem with using `InlineString`s.  the fix requires changes to both Compose and Gadfly.  PRs are in the works.

bjarthur:
RDatasets uses `String7` now.  that's the source of the problem.

bjarthur:
here's the fixes to Compose:  https://github.com/GiovineItalia/Compose.jl/pull/438

bjarthur:
and here's the fixes to Gadfly https://github.com/GiovineItalia/Gadfly.jl/pull/1608

PR #1608, opened 2023-02-20
fix for AbstractString other than String, and doc example

bjarthur:
fixes https://github.com/GiovineItalia/Gadfly.jl/issues/1607


bjarthur:
tests won't pass until https://github.com/GiovineItalia/Compose.jl/pull/438 is marged and released

PR #1617, opened 2023-06-20
bump minor version

bjarthur:
@Mattriks @tlnagy main impetus for doing this is that gadfly is holding up upgrading many of it's dependents.  there are also a few new features, hence bumping the minor version instead of the patch.  you two okay with merging this?


tlnagy:
Hi @bjarthur, I don't have time to work on Gadfly for the foreseeable future so I'm okay with whatever you and @Mattriks want to do.

PR #1638, opened 2025-03-10
fix world age errors in tests

KristofferC:
There is some difference in how world ages are handled on 1.12 and this evals a file inside a block so the definitions in there are in general not visible to code in the same block where the world age has not gotten the chance to be bumped. This makes the tests pass on 1.12 again.


KristofferC:
CI is broken

KristofferC:
New release with this would fix PkgEval for this package.

bjarthur:
https://github.com/JuliaRegistries/General/pull/132565

